PR #7, opened 2017-03-25
Miscellaneous improvements

bluetech:
Some changes I made while playing with this library. Please see the commit messages.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ base64 = "0.4.0"
 byteorder = "1.0.0"
 bytes = "0.4.1"
 httparse = "1.2.1"
-env_logger = "0.4.2"
 log = "0.3.7"
 rand = "0.3.15"
 sha1 = "0.2.0"
@@ -30,3 +29,6 @@ utf-8 = "0.7.0"
 [dependencies.native-tls]
 optional = true
 version = "0.1.1"
+
+[dev-dependencies]
+env_logger = "0.4.2"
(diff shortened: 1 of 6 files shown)


bluetech:
I got around to running the autobahn test suite. This PR introduces regressions. I will fix them when I have time and report back.

bluetech:
OK, the autobahn test suite is green again. I removed a couple of commits. I can see why you went to the trouble with the `StringCollector` now - autobahn really wants us to fail fast on invalid UTF-8.

Once https://github.com/rust-lang/rust/pull/40212 stabilizes, it should be possible to drop the `utf-8` crate and use `str::from_utf8()` again.

agalakhov:
Thank you! Looks good. I'll test and merge it ASAP. You're mostly improving the code I copied from ws-rs to speed-up the development, and it really needed a cleanup.

We have the same "early fail on invalid UTF-8" issue with frames too. That's why two tests are still non-strict. There is a possible way to fix it by letting the frame parser return incomplete frames.

bluetech:
I added another commit fixing a problem with `InputBuffer` (tell me if this annoys you, and I will open a new PR instead).

ffablia:
This PR seems to resolve an infinite memory consumption with high throughput.

I am using throughput benchmark (./throughput 10000 20 10 3000) from uWebSockets against twists echo server and a mio/tungstenite echo server.
Without this PR, both version are using more and more memory.
With this PR, both are using constant memory size.

PR #17, opened 2017-07-20
Add `Ping` and `Pong` variants to `Message`

### LINKED ISSUE #16: Handling "pong" messages
Hello! I was tinkering around and wondering how to implement a timeout of a pong message. That is, after a ping is sent, having a maximum amount of time before the pong is received after which the connection is torn down. It looks like right now though once a pong message is received currently [no action is taken](https://github.com/snapview/tungstenite-rs/blob/9d6c3058d5dadbf3be86b8cb16d9da0b4e401a00/src/protocol/mod.rs#L367-L375). 

I was wondering, is there perhaps another method to currently implement this? If not, do y'all have thoughts on how this might be implemented?

alexcrichton:
This'll allow removal/deprecation of the `send_ping` custom method but also
allow applications to react to `Pong` messages as appropriate.

Closes #16

### DIFF
--- a/examples/autobahn-client.rs
+++ b/examples/autobahn-client.rs
@@ -5,7 +5,7 @@ extern crate url;
 
 use url::Url;
 
-use tungstenite::{connect, Error, Result};
+use tungstenite::{connect, Error, Result, Message};
 
 const AGENT: &'static str = "Tungstenite";
 
@@ -33,8 +33,14 @@ fn run_test(case: u32) -> Result<()> {
     ).unwrap();
     let mut socket = connect(case_url)?;
     loop {
-        let msg = socket.read_message()?;
-        socket.write_message(msg)?;
+        match socket.read_message()? {
+            msg @ Message::Text(_) |
+            msg @ Message::Binary(_) => {
+                socket.write_message(msg)?;
+            }
+            Message::Ping(_) |
+            Message::Pong(_) => {}
+        }
     }
 }
 
--- a/examples/autobahn-server.rs
+++ b/examples/autobahn-server.rs
@@ -5,7 +5,7 @@ extern crate tungstenite;
 use std::net::{TcpListener, TcpStream};
 use std::thread::spawn;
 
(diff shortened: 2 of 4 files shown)


application-developer-DA:
Thanks for the pull request.

Seems good, the only remarks/questions are:

1. Considering the pull request changes, tungstenite will not schedule a pong response to the ping message automatically anymore, the user has to react on the ping message with a pong message, I don't know if it is a good thing or not, as it leads to a boilerplate in a user code (the user has to react with a pong message after receiving a ping); we anyway have to reply with a pong message whenever we receive a ping message according to the RFC, so would it still makes sense to send a pong message after the ping message is received inside the tungstenite? (as it was done before)
2. It would be nice to change the examples section accordingly, if I merge this pull request some of the autobahn-tests from examples section will fail (the autobahn examples should reply properly on ping/pong messages, the current state of examples does not expect that the `Message` enum contains anything other than `Binary` or `Text` messages and so it will reply with a send `Ping` and `Pong` messages back to the sender, when it receives something).

Other than that, it looks nice. I can merge the changes, but I can't publish the crate (do you plan to rely on a published version), usually @agalakhov does this (he has an account and credentials there) and also he might want to have a look at the changes before publishing (currently he's in vacation, so he'll check it either if he finds some time during the vacation or when he's back in the office).

alexcrichton:
> so would it still makes sense to send a pong message after the ping message is received inside the tungstenite? 

Certainly!

> if I merge this pull request some of the autobahn-tests from examples section will fail 

Oh is there a way to run these locally to verify the changes as well?

> do you plan to rely on a published version

Ah no urgent need! A git dependency for now is fine.

alexcrichton:
Ok I think I managed to figured it out, hopefully green now!

application-developer-DA:
> Ah no urgent need! A git dependency for now is fine.

FYI: Both tungstenite and tokio-tungstenite has just been published by @agalakhov , they contain the changes from your pull request + recent changes in regards to HTTP headers support.

Issue #22, opened 2017-09-13
Custom buffer management


alexcrichton:
In the server I've been working on the load profile of it currently has thousands/millions of connected websockets, but they're all idle. One particular constraint has been memory so far, and we're thinking that a possible optimization would be to omit the `InputBuffer` for idle connections. In theory with an async server there's no need to actually allocate a buffer for input until the socket is itself readable.

I was curious if y'all had thought about this before? I think this could possible get built externally with just `Frame` and other low-level bits, but ideally we could reuse the `WebSocket` logic itself without having to reimplement various portions.

agalakhov:
I wanted to avoid extra allocations here. In our case of all active connections on the server allocating would be too expensive. Also, having the data preallocated lowers the risk of being killed by the OOM killer. So my choice is to make it configurable. In fact, there are exactly two configuration values: the maximum buffer size and the minimum allocation amount (that's constantly kept allocated). Zero amount means no preallocation then.

alexcrichton:
@agalakhov that makes sense to me! I wonder if perhaps something like this could be done?

```rust
pub trait WebsocketIo: Read  + Write {
    fn is_readable(&self) -> bool;
    fn max_read_buffer_size(&self) -> usize;
    fn min_read_buffer_size(&self) -> usize;
    fn max_write_buffer_size(&self) -> usize;
    fn min_write_buffer_size(&self) -> usize;
}
```

Most internal functions could be generic over `WebsocketIo` where top-level functions might look like:

```rust
pub fn accept<S: Read + Write>(stream: S, callback: Option<Callback>)
    -> Result<WebSocket<S>, HandshakeError<ServerHandshake<S>>>
{
    accept_io(DefaultIo(stream), callback)
}

pub fn accept_io<S: WebsocketIo>(stream: S, callback: Option<Callback>)
    -> Result<WebSocket<S>, HandshakeError<ServerHandshake<S>>>
{
    // implementation ...
}
```

where basically all `Read + Write` objects still get reasonable defaults, but there's a vector to insert a custom trait implementation as well?

agalakhov:
Good idea, makes sense! Thanks!

PR #26, opened 2018-03-24
`WebSocket::write_message` only buffers one message at a time to apply back-pressure

tedsta:
I figured if anyone wants to buffer up messages, they can just use `Sink::buffer()`. If you like the idea of a configurable size `send_queue` I can do that instead.

Oh also I replaced some `replace()` (heh) calls with `take()`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,9 +7,9 @@ authors = ["Alexey Galakhov"]
 license = "MIT/Apache-2.0"
 readme = "README.md"
 homepage = "https://github.com/snapview/tungstenite-rs"
-documentation = "https://docs.rs/tungstenite/0.5.3"
+documentation = "https://docs.rs/tungstenite/0.5.4"
 repository = "https://github.com/snapview/tungstenite-rs"
-version = "0.5.3"
+version = "0.5.4"
 
 [features]
 default = ["tls"]
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -42,6 +42,7 @@ pub struct WebSocket<Stream> {
     incomplete: Option<IncompleteMessage>,
     /// Send: a data send queue.
     send_queue: VecDeque<Frame>,
+    max_send_queue: usize,
     /// Send: an OOB pong message.
     pong: Option<Frame>,
 }
@@ -74,6 +75,7 @@ impl<Stream> WebSocket<Stream> {
             state: WebSocketState::Active,
             incomplete: None,
             send_queue: VecDeque::new(),
+            max_send_queue: 1,
             pong: None,
         }
     }
@@ -101,13 +103,20 @@ impl<Stream: Read + Write> WebSocket<Stream> {
 
     /// Send a message to stream, if possible.
     ///
-    /// This function guarantees that the frame is queued regardless of any errors.
-    /// There is no need to resend the frame. In order to handle WouldBlock or Incomplete,
-    /// call write_pending() afterwards.
+    /// WebSocket will buffer a configurable number of messages at a time, except to reply to Ping
+    /// and Close requests. If the WebSocket's send queue is full, SendQueueFull will be returned
+    /// along with the passed message. Otherwise, the message is queued and Ok(()) is returned.
     ///
     /// Note that only the last pong frame is stored to be sent, and only the
-    /// most recent pong frame is sent if multiple pong frames are queued up.
+    /// most recent pong frame is sent if multiple pong frames are queued
     pub fn write_message(&mut self, message: Message) -> Result<()> {
+        // Try to make some room for the new message
+        self.write_pending().no_block()?;
+
+        if self.send_queue.len() >= self.max_send_queue {
+            return Err(Error::SendQueueFull(message));
+        }
+
         let frame = match message {
             Message::Text(data) => {
                 Frame::message(data.into(), OpCode::Data(OpData::Text), true)
@@ -122,13 +131,13 @@ impl<Stream: Read + Write> WebSocket<Stream> {
             }
         };
         self.send_queue.push_back(frame);
-        self.write_pending()
+        Ok(())
     }
 
     /// Close the connection.
     ///
     /// This function guarantees that the close frame will be queued.
-    /// There is no need to call it again, just like write_message().
+    /// There is no need to call it again.
     pub fn close(&mut self, code: Option<CloseFrame>) -> Result<()> {
         if let WebSocketState::Active = self.state {
             self.state = WebSocketState::ClosedByUs;
@@ -151,7 +160,7 @@ impl<Stream: Read + Write> WebSocket<Stream> {
         // Upon receipt of a Ping frame, an endpoint MUST send a Pong frame in
         // response, unless it already received a Close frame. It SHOULD
         // respond with Pong frame as soon as is practical. (RFC 6455)
-        if let Some(pong) = replace(&mut self.pong, None) {
+        if let Some(pong) = self.pong.take() {
             self.send_one_frame(pong)?;
         }
         // If we have any unsent frames, send them.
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -7,9 +7,9 @@ authors = ["Alexey Galakhov"]
 license = "MIT/Apache-2.0"
 readme = "README.md"
 homepage = "https://github.com/snapview/tungstenite-rs"
-documentation = "https://docs.rs/tungstenite/0.5.3"
+documentation = "https://docs.rs/tungstenite/0.5.4"
 repository = "https://github.com/snapview/tungstenite-rs"
-version = "0.5.3"
+version = "0.5.4"
```
application-developer-DA: The changes in this PR are not backward-compatible with the previous version (see the error types returned by some of the functions for instance), so I think we cannot just change a patch version in this case according to Semantic Versioning paradigm.

src/protocol/mod.rs
```diff
@@ -101,13 +102,16 @@ impl<Stream: Read + Write> WebSocket<Stream> {
...
+    /// WebSocket will only buffer one message at a time, except to reply to Ping and Close
+    /// requests. If the WebSocket is still busy with the previous message, WouldBlock will be
+    /// returned along with the passed message. If after the passed message is buffered, an error
+    /// occurs (including WouldBlock), the error will be returned with no message.
     ///
-    /// Note that only the last pong frame is stored to be sent, and only the
-    /// most recent pong frame is sent if multiple pong frames are queued up.
-    pub fn write_message(&mut self, message: Message) -> Result<()> {
+    /// In order to handle WouldBlock or Incomplete, call write_pending() afterwards.
+    pub fn write_message(&mut self, message: Message) -> result::Result<(), (Error, Option<Message>)> {
```
application-developer-DA: Oh, well, I think it will be harder to integrate this semantics to the modern Tokio state.

`write_message()` were meant to have the same responsibilities as former `Sink::start_send()` (NOTE: the semantics of `start_send()` change in newer `futures-rs` versions), i.e. it simply either pushes the message in the queue or (in case if we implement a backpressure), returns the message back when the queue is full.

application-developer-DA:
Thanks for the pull request @tedsta , but based on our short discussion with @agalakhov in another issue I believe that the expectations are that the backpressure should be supported directly by `tungstenite-rs`.

Also it seems that these changes will make the usage of `tungstenite-rs` less obvious and more verbose (as you cannot specify "I just want to have an unlimited queue in order to be able to push messages to it"), and it forces the users of `tungstenite-rs` implement some wrappers to have some sort of buffering queue in case if they want to buffer `N` messages (also it does not seem to solve the problem of backpressure as such). So I think in this manner it's even less usable than the current state (with the current implementation you can implement a backpressure on top of `tungstenite-rs` easily by for example using the `buffer()` method which you mentioned before).

It also seems that the new semantics is less "compatible" with one of the suggestions from Alex Crichton on this topic: https://github.com/snapview/tungstenite-rs/issues/22#issuecomment-329619966


> I figured if anyone wants to buffer up messages, they can just use Sink::buffer().

I assume you're talking about `SinkExt::buffer()`? Yes, that makes sense for me. Actually the current semantics for `Sink` has changed. If you look at the futures `0.2.0`, the semantics of `start_send()` has changed. Now it does not necessary has to return the message if it could not push it to the queue, moreover, now the error on `start_send()` has completely different meaning. Probably we have to rethink how to add the backpressure properly.

tedsta:
TLDR: I think that the current behavior is alright, but *I still feel that backpressure by default is better*, and these changes only trivially affect the migration to futures 0.2. I'm also not super invested in this PR since I've already worked around it.

> Also it seems that these changes will make the usage of tungstenite-rs less obvious and more verbose (as you cannot specify "I just want to have an unlimited queue in order to be able to push messages to it")

Regarding "less obvious": I disagree - if you look to the standard library, `std::net::TcpStream` will block, and `std::sync::SyncSender` will return the message back if it can't be sent yet. Tokio's provided Sinks mirror this behavior. So I expect backpressure by default. If you're using Tokio, you can simply `Stream::forward` messages from an unbounded channel to the `WebSocketStream` if you really need an unlimited queue.

Regarding "more verbose": I agree, but I also feel this is necessary. I do think `(Error, Option<Message>)` is ugly and should probably be wrapped in a struct. It is analogous `std::sync::mpsc::SendError`, except lucky for `std::sync::mpsc::SyncSender`, sends are either completely sent all at once or not at all. `WebSocket`s *have* to allow partial progress in an async setting because it's not guaranteed that the underlying socket's write buffer will be large enough to hold an entire `Message`, so you have the following possible situations:

1. Unable to start sending the message - return `Error` and the `Message`
2. Able to start sending but unable to flush - return `Error` but no `Message`
3. Able to both send and flush - return nothing

And unfortunately which `Error` occurred matters here, so we have to return it along with the `Message`.

> forces the users of tungstenite-rs implement some wrappers to have some sort of buffering queue in case if they want to buffer N message

I haven't tested this, but since this crate uses `std::net::TcpStream` by default, I think `WebSocket::write_message()` will block if the underlying socket's write buffer is full (which is good IMO). To handle buffering, I think it's fairly easy to do something like:

```rust
let (ws_tx_tx, ws_tx_rx) = sync::mpsc::sync_channel(10);
thread::spawn(move || {
    while let Ok(msg) = ws_tx_rx.recv() {
        if let Err(e) = ws_stream.write_message(msg) {
            // Handle error
        }
    }
});
```

And if you need an unlimited buffer, you can use `std::sync::mpsc::channel()` instead. Note that you don't have to call `WebSocket::write_pending` in this case because `WebSocket::write_message` should flush the entire message so long as blocking is enabled on the underlying socket.

> with the current implementation you can implement a backpressure on top of tungstenite-rs easily by for example using the buffer() method which you mentioned before

This isn't true, because the buffer will always instantly flush any message it gets via `Sink::start_send` to `WebSocketStream`, because `WebSocketStream::start_send` will never turn down a message.

Sure, you can implement backpressure on top of tungstenite-rs by always ensuring `write_pending` doesn't return `WouldBlock`, but:
1. Doing so feels backwards - when using Tokio, it is trivial to add buffering to a `Sink` that applies backpressure (by using `Sink::buffer()`). You need a custom `Sink` wrapper to add back-pressure to a `Sink` that doesn't have it.
2. In our discussion in snapview/tokio-tungstenite#35, you guys decided the correct place for back-pressure is in tungstenite-rs, hence the PR is here.

> I assume you're talking about SinkExt::buffer()

I'm talking about `Sink::buffer()` in futures 0.1. Since tokio-tungstenite hasn't been ported to 0.2 yet, IMO it's better to discuss this in terms of futures 0.1, of course with consideration for how it will affect conversion to futures 0.2.

> Now it does not necessary has to return the message if it could not push it to the queue, moreover, now the error on start_send() has completely different meaning

I think "completely different meaning" is a bit strong. The only difference I see actually makes it easier on `Sink::start_send` implementers, since callers *must* successfully call `Sink::poll_ready` before calling it.

If you guys have changed your minds about tungstenite-rs being the correct place for backpressure, it is pretty trivial to implement backpressure in tokio-tungstenite with the current behavior of tungstenite-rs, regardless of which version of futures is used:

- For futures 0.1, simply attempt `WebSocket::write_pending()` before attempting `WebSocket::write_message()`.
- For futures 0.2, `<WebSocketStream as Sink>::poll_ready` and `<WebSocketStream as Sink>::poll_flush` both simply attempt `WebSocket::write_pending`. `<WebSocketStream as Sink>::start_send` just attempts `WebSocket::write_message`.

With the behavior from this PR, the implementation is roughly the same (just slightly different error handling in `impl Sink for WebSocketStream`, the only difference is that `<WebSocketStream as Sink>::start_send` will fail if it hasn't been flushed before calling, which fits the contract: https://docs.rs/futures/0.2.0-beta/futures/sink/trait.Sink.html?search=#tymethod.start_send

In any case, I've already worked around the lack of backpressure, so I'm not super invested in this PR. I am willing to put a little more work in to get it merged if you do want it though.

application-developer-DA:
> I think that the current behavior is alright, but I still feel that backpressure by default is better, and these changes only trivially affect the migration to futures 0.2. I'm also not super invested in this PR since I've already worked around it.

Well, in general I got your point and we still think that `tungstenite-rs` is a better place to implement this sort of a backpressure, however I would implement it a bit different. Initially we expected that the user would be able to configure the length of the queue, i.e. the one can choose between "unlimited queue" (the current approach) and "limited queue" behavior. When the amount of elements in the queue exceeds the specified maximum queue size, then it behaves the same way `Sink::start_send()`, i.e. it will return an element. If you could update the implementation a bit, so that  `write_message()` reflects its semantics and wrap the type which returns `write_message()`, then it would match the expectations which we initially had. It's pefectly fine if you don't have time to work on it anymore or made a workaround in a different way (as you mentioned in your comment), we'll try to implement it as soon as possible.

-----

TLDR:

> Regarding "less obvious": I disagree - if you look to the standard library, std::net::TcpStream will block, and std::sync::SyncSender will return the message back if it can't be sent yet.

That's true, but both of them operate on sync abstractions, i.e. `std::net::TcpStream` is a blocking tcp stream. Imagine that you have a mio-based framework on top of the `tungstenite-rs` which relies on `tungstenite-rs` to keep the queue and manage it, making the `tungstenite-rs` to accept only 1 message at time makes your framework responsible for handling message queues, we are awared about one of such frameworks already, which works in a way similar to `ws-rs` (it's like `ws-rs`, but uses `tunstenite-rs`), hardcoding a queue size to be equal to one would result in a new version of `tungstenite-rs` which would be incompatible with the existing framework, moreover the framework does not want to handle the queues, that's why [ideally] we would like to make the queue size configurable yet flexible enough to make it compatible with the `tokio-tungstenite` (in this case we may want to set the queue size to 1 when the websocket is created from `tokio-tungstenite`).

tedsta:
> Imagine that you have a mio-based framework on top of the tungstenite-rs which relies on tungstenite-rs to keep the queue and manage it, making the tungstenite-rs to accept only 1 message at time makes your framework responsible for handling message queues

Ah, I hadn't considered using this with mio. I still feel that the user should be responsible for managing queues, just as tokio takes up that responsibility, but I don't feel strongly about it.

I think I implemented the changes you requested. Let me know what you think.

application-developer-DA:
Looks good!

> Ah, I hadn't considered using this with mio. I still feel that the user should be responsible for managing queues, just as tokio takes up that responsibility, but I don't feel strongly about it.

There are some server side projects in production which are using `tungstenite-rs` + `mio`, some of them rely on the fact that `tungstenite-rs` manages the queue internally.

> EDIT: Oh, one thing I haven't implemented yet is how to actually configure the max queue size (it's hard-coded right now). Do you want to make the current methods default to some huge number and provide new methods that also allow specifying max queue size? That would maintain reasonable compatibility. I guess the minor version number would have to be bumped anyway because of the new SendQueueFull error enum variant though.

Right. I think ideally the one has to be able to specify `max_queue_size: Option<usize>` where `None` would mean "infinite" queue (like it is done in the current implementation in master).

agalakhov:
If, for some reason, you don't want to use `None`, `<usize>::max_value()` would do the same.

application-developer-DA:
Ok, since neither me nor @agalakhov have objections (apart small things which may be improved the pull request), I'll merge the pull request and we'll implemented the suggested improvements on top of the master branch. @tedsta thanks for adding the suppotr for the backpressure!

PR #37, opened 2018-07-10
frame: eliminate an unsafe when reading message data into the buffer

bluetech:
Since there is only one `unsafe` left outside of `frame/mask.rs`, I thought it'd be fun to try and get rid of it.

The reason for the unsafe was presumably to avoid copying or zero-initializing. This achieves the same but using only safe functions.

Note: there is no zero-initializing here because `Cursor` is "trusted" to not read from the buffer and so skips the initialization: https://github.com/rust-lang/rust/blob/master/src/libstd/io/cursor.rs#L241 (the `Take` wrapper delegates to its inner).


agalakhov:
Thank you!

PR #43, opened 2018-07-16
Limiting and config improvements

agalakhov:


src/protocol/frame/frame.rs
```diff
@@ -254,86 +243,34 @@ impl Frame {
...
+    /// masked. In other words, those frames that have just been received from a client endpoint.
     #[inline]
-    pub fn remove_mask(&mut self) {
-        self.header.mask.and_then(|mask| {
-            Some(apply_mask(&mut self.payload, &mask))
-        });
-        self.header.mask = None;
+    pub(crate) fn apply_mask(&mut self) {
+        if let Some(mask) = self.header.mask.take() {
+            apply_mask(&mut self.payload, &mask)
```
application-developer-DA: In the previous version the mask had been set to `None`, now it does not happen. Is it an expected behavior?

src/protocol/mod.rs
```diff
@@ -108,6 +108,11 @@ impl<Stream> WebSocket<Stream> {
     pub fn get_mut(&mut self) -> &mut Stream {
         self.socket.get_mut()
     }
+
+    /// Change the configuration.
+    pub fn set_config(&mut self, set_func: impl FnOnce(&mut WebSocketConfig)) {
+        set_func(&mut self.config)
```
application-developer-DA: Hm, but it could be misleading as the one may thing that modifying the configuration will immediately affect the websocket connection, which in fact would not happen (the internal `SocketFrame` is not modified).

I think that the configuration for a single websocket connection once on initialization should be sufficient, isn't it?

agalakhov: `FrameSocket` does not have any internal configuration state, all parameters are passed at function calls. This `self.config` is the only place where configuration is stored and applied. Thus it takes effect immediately.

PR #44, opened 2018-07-16
Add fuzzing harness and seeds

Shnatsel:
This PR adds a fuzzing harness via cargo-fuzz, as well as the minified corpora of input files generated by the fuzzer.

The fuzzing harness contains code copypasted from tungstenite itself, which should be fixed, but I'm not sure how to go about it.

The most interesting part is probably the generated corpora. I have obtained them by fuzzing the targets with afl so that performance would not be degraded by sanitizers in cargo-fuzz, then using cargo-fuzz for further validation and minifying the corpus for each target. All of this has found no crashes. Out of all crates I've applied AFL to, tungstenite is the first crate to withstand afl without a single panic. Kudos!

I imagine this will need a revision after #43 lands, but I'm opening a PR now because not sure I'll be able to complete this.


bluetech:
I have run each target for a few minutes; nice!

One thing I've noticed is that the RSS is quickly increasing without bound. This is not supposed to happen, I think. You should try to investigate what is causing it.

Shnatsel:
It stops increasing when it reaches 500Mb or so per process. This is a side effect of using address sanitizer. Using at least one sanitizer is currently mandatory in cargo-fuzz. If there was a memory leak, cargo-fuzz would have detected it.

bluetech:
Uh, makes sense. I see that address sanitizer has a default 256mb "quarantine" ([here](https://github.com/google/sanitizers/wiki/AddressSanitizerFlags#run-time-flags)).

PR #52, opened 2018-12-19
Nonblocking/timeout-compatible reads.

FelixMcFelix:
Adds a new method to a `WebSocket`: `try_read_message`.
Calling `read_message` would always enter a loop until receipt of a
message, invalidating any modifications to the underlying stream's
timeout or blocking mode. The new method allows these to be respected.

This is structured so that the semantics/behaviour of `read_message`
remain unchanged, exposing the inner part of its loop publicly.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -141,17 +141,38 @@ impl<Stream: Read + Write> WebSocket<Stream> {
     /// However, it never blocks on write.
     pub fn read_message(&mut self) -> Result<Message> {
         loop {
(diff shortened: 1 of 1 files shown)


agalakhov:
Are you sure it really invalidates notifications? We actually run tungstenite in its non-blocking mode in production, and it does not wait for messages in a loop.

FelixMcFelix:
It does work as you say, the issues on my end arose from how `Err`s/`None`s were being handled that caused the `WouldBlock` to go unnoticed. Closing.

PR #53, opened 2019-01-14
Upgrade rand 0.5.2 -> 0.6.4

resilar:
Tokio & futures 0.1 ecosystem seems to have moved to rand 0.6 except for Tungstenite. At least in my [toyproject](https://github.com/resilar/kuppalista), Tungstenite is the only dependency that still depends on rand 0.5.2. This patch resolves the only remaining duplicate transitive dependency in my project (since no two rand versions are required after applying this patch).


application-developer-DA:
Thanks.

PR #55, opened 2019-01-25
Add support for HTTP proxies with optional Basic Authentication

bschwind:
I wrote a CLI application which is sometimes used behind corporate proxies. Although the HTTP client I'm using supports configuring a proxy, this library did not, so here I am.

It currently only supports HTTP proxies (not HTTPS or SOCKS though I imagine those would be fairly easy to add) and can optionally use Basic Auth for authentication.

Halfway through implementing, I realized there's a `client` function which allows you to pass in your own `Read + Write` stream. I tried using it but found myself reimplementing a lot of what already exists inside this crate, especially the `connect_to_some` and `wrap_stream` functions.

If you feel this functionality is inappropriate to this crate and want to close this PR, that's totally okay! In that case I'll use our fork and be done with it. It just seemed like no WebSocket clients support this in an easy fashion. However I'm pretty new to interacting with proxy servers so perhaps there's a more obvious way to do what I want in these libraries (tungstenite, websocket-rs, etc.) without this change. If so, please let me know!

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,12 +1,15 @@
 //! Methods to connect to an WebSocket as a client.
 
-use std::net::{TcpStream, SocketAddr, ToSocketAddrs};
+use std::net::{TcpStream, ToSocketAddrs};
 use std::result::Result as StdResult;
 use std::io::{Read, Write};
 
 use url::Url;
 
-use handshake::client::Response;
+use httparse;
+use httparse::Status;
+
+use handshake::client::{Response, ProxyAuth};
 use protocol::WebSocketConfig;
 
 #[cfg(feature="tls")]
@@ -87,8 +90,8 @@ pub fn connect_with_config<'t, Req: Into<Request<'t>>>(
 ) -> Result<(WebSocket<AutoStream>, Response)> {
     let request: Request = request.into();
     let mode = url_mode(&request.url)?;
-    let addrs = request.url.to_socket_addrs()?;
-    let mut stream = connect_to_some(addrs, &request.url, mode)?;
+    // let addrs = request.url.to_socket_addrs()?;
+    let mut stream = connect_to_some(&request, mode)?;
     NoDelay::set_nodelay(&mut stream, true)?;
     client_with_config(request, stream, config)
         .map_err(|e| match e {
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -115,19 +117,91 @@ pub fn connect<'t, Req: Into<Request<'t>>>(request: Req)
...
+                        ProxyAuth::Basic(basic_auth) => {
+                            buf.extend_from_slice(&format!("Proxy-Authorization: {}\r\n", basic_auth.to_header_value()).as_bytes());
+                        }
+                    }
+                }
+
+                // Add the trailing empty line
+                buf.extend_from_slice(b"\r\n");
+
+                let _ = raw_stream.write(&buf)?;
```
bschwind: I might need to continue in the `for addr in addrs` loop instead of early -returning here.

src/client.rs
```diff
@@ -115,19 +117,91 @@ pub fn connect<'t, Req: Into<Request<'t>>>(request: Req)
...
+
+                let mut read_buf = [0; 256];
+                let mut response_vec = Vec::with_capacity(1024);
+
+                loop {
+                    let n = raw_stream.read(&mut read_buf)?;
```
bschwind: Same here

src/client.rs
```diff
@@ -115,19 +117,91 @@ pub fn connect<'t, Req: Into<Request<'t>>>(request: Req)
...
+
+                    response_vec.extend_from_slice(&read_buf[..n]);
+
+                    let mut hbuffer = [httparse::EMPTY_HEADER; super::handshake::headers::MAX_HEADERS];
+                    let mut res = httparse::Response::new(&mut hbuffer);
+
+                    match res.parse(&response_vec)? {
```
bschwind: And here

bschwind:
> Our approach is to negotiate proxy connection separately and pass a socket to Tungstenite then

That makes sense! As I mentioned, I wanted to do it that way but I felt like I was replicating a lot of the original tungstenite logic. Feel free to close this, but I would be very interested to hear if you ever open source the proxy support crate you mentioned.

application-developer-DA:
I'm currently checking some PRs to see which ones can be closed / merged. This one seems to be obsolete and conflicts with master. I'll close it, because it actually requires some deeper changes, i.e. "open-sourcing" our proxy support library and/or using some other libraries for that, so that `tungstnite-rs` does not need to deal with non-websocket-related things as discussed in previous comments.

metaclips:
Hello, any update on the proxy support library @application-developer-DA

application-developer-DA:
@metaclips unfortunately no updates here. We do have a proxy library which we have not open-sourced yet (it still requires some clean ups etc), we could theoretically publish it "as is" though.

metaclips:
Please do. Thanks.

PR #58, opened 2019-03-23
extract the context values from `WebSocket`

ubnt-intrepid:
This PR extracts the implementation details in `protocol::WebSocket<S>` as a new struct `WebSocketContext`.

The motivation of this change is to allow handling WebSocket data when the provided I/O is borrowed and cannot be create the instance of `WebSocket`.

### DIFF
--- a/src/protocol/frame/mod.rs
+++ b/src/protocol/frame/mod.rs
@@ -18,12 +18,8 @@ use error::{Error, Result};
 pub struct FrameSocket<Stream> {
     /// The underlying network stream.
     stream: Stream,
-    /// Buffer to read data from the stream.
-    in_buffer: InputBuffer,
-    /// Buffer to send packets to the network.
-    out_buffer: Vec<u8>,
-    /// Header and remaining size of the incoming packet being processed.
-    header: Option<(FrameHeader, u64)>,
+    /// Codec for reading/writing frames.
+    codec: FrameCodec,
 }
 
 impl<Stream> FrameSocket<Stream> {
@@ -31,9 +27,7 @@ impl<Stream> FrameSocket<Stream> {
     pub fn new(stream: Stream) -> Self {
         FrameSocket {
             stream,
-            in_buffer: InputBuffer::with_capacity(MIN_READ),
-            out_buffer: Vec::new(),
-            header: None,
+            codec: FrameCodec::new(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/protocol/frame/mod.rs
```diff
@@ -18,38 +18,30 @@ use error::{Error, Result};
 pub struct FrameSocket<Stream> {
     /// The underlying network stream.
     stream: Stream,
-    /// Buffer to read data from the stream.
-    in_buffer: InputBuffer,
-    /// Buffer to send packets to the network.
-    out_buffer: Vec<u8>,
-    /// Header and remaining size of the incoming packet being processed.
-    header: Option<(FrameHeader, u64)>,
+    /// Codec for reading/writeing frames.
```
application-developer-DA:
```suggestion
    /// Codec for reading/writing frames.
```

src/protocol/frame/mod.rs
```diff
@@ -68,6 +60,63 @@ impl<Stream> FrameSocket<Stream>
...
+    pub(super) fn from_partially_read(part: Vec<u8>) -> Self {
+        Self {
+            in_buffer: InputBuffer::from_partially_read(part),
+            out_buffer: Vec::new(),
+            header: None,
+        }
+    }
+
+    /// Read a frame from the provided stream.
+    pub(super) fn read_frame<Stream>(&mut self, stream: &mut Stream, max_size: Option<usize>) -> Result<Option<Frame>>
```
application-developer-DA: Does it make sense to reformat this line, so that it's a bit more readable?

(I believe it's quite long)

src/protocol/mod.rs
```diff
@@ -97,77 +90,153 @@ impl<Stream> WebSocket<Stream> {
...
+    /// This function guarantees that the close frame will be queued.
+    /// There is no need to call it again. Calling this function is
+    /// the same as calling `write(Message::Close(..))`.
+    pub fn close(&mut self, code: Option<CloseFrame>) -> Result<()> {
+        self.context.close(&mut self.socket, code)
+    }
+}
+
+
+/// A context for managing an WebSocket stream.
```
application-developer-DA:
```suggestion
/// A context for managing WebSocket stream.
```

src/protocol/mod.rs
```diff
@@ -97,77 +90,153 @@ impl<Stream> WebSocket<Stream> {
...
+    /// Send: a data send queue.
+    send_queue: VecDeque<Frame>,
+    /// Send: an OOB pong message.
+    pong: Option<Frame>,
+    /// The configuration for the websocket session.
+    config: WebSocketConfig,
+}
+
+impl WebSocketContext {
+    /// Create a WebSocket context that manages an post-handshake stream.
```
application-developer-DA:
```suggestion
    /// Create a WebSocket context that manages a post-handshake stream.
```

ubnt-intrepid:
>  the main goal is to make it possible to use `WebSocketContext` to just "parse" and "send" websocket messages into some abstract stream that supports `Read` and `Write`

Yes, it is correct.

PR #59, opened 2019-05-04
server: let callback return HTTP error messages

agalakhov:
Signed-off-by: Alexey Galakhov <agalakhov@snapview.de>

### DIFF
--- a/src/handshake/server.rs
+++ b/src/handshake/server.rs
@@ -3,9 +3,11 @@
 use std::fmt::Write as FmtWrite;
 use std::io::{Read, Write};
 use std::marker::PhantomData;
+use std::result::Result as StdResult;
 
 use httparse;
 use httparse::Status;
+use http::StatusCode;
 
 use error::{Error, Result};
 use protocol::{WebSocket, WebSocketConfig, Role};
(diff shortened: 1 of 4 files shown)


src/handshake/server.rs
```diff
@@ -141,24 +179,48 @@ impl<S: Read + Write, C: Callback> HandshakeRole for ServerHandshake<S, C> {
                 if !tail.is_empty() {
                     return Err(Error::Protocol("Junk after client request".into()))
                 }
-                let extra_headers = {
-                    if let Some(callback) = self.callback.take() {
-                        callback.on_request(&result)?
```
application-developer-DA: But the error in the previous version also was returned, right? So the biggest difference is that the type of the error has been changed? And so does the handling, i.e. instead of returning an error, the function will rreturn `ProcessingResult::Continue` and write some response to the incoming request, correct?

agalakhov: The previous version had wrong error type. It used the same `Error` as the library itself which is nonsense. Returning an error from the callback resulted in immediate connection drop and returning the same error to the library user.

PR #60, opened 2019-05-14
Fixes for the "close-is-a-message" semantics.

agalakhov:
`Message::Close` corresponds to a "close" message received from peer.
Connection close corresponds to `Error::ConnectionClosed` but payload is removed.
Trying to use closed connection corresponds to `Error::AlreadyClosed`.


application-developer-DA:
We should also re-run autobahn tests (better would be to add them into our Travis), just to make sure that everything is ok.

agalakhov:
I already checked it with autobahn in both client and server mode.

PR #63, opened 2019-07-24
Update the url crate to 2.0

SimonSapin:


SimonSapin:
Could you publish on crates.io? Thanks!

PR #72, opened 2019-09-11
Keep processing incoming data even after we have initiated a close handshake.

najamelan:
This commit will make it so that after initiating a close handshake but before receiving confirmation from the remote. An integration test is included verifying that it works correctly.

This fixes the issue as described here: https://github.com/snapview/tokio-tungstenite/issues/69#issuecomment-530592464

Note that the websocket RFC is not specific about this. There is however no good reason for tungstenite to impose unnecessary data loss on clients.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -396,7 +396,7 @@ impl WebSocketContext {
                     }
                 }
 
-                OpCode::Data(_) if !self.state.is_active() => {
+                OpCode::Data(_) if !self.state.can_read() => {
                     // No data processing while closing.
                     Ok(None)
                 }
@@ -552,6 +552,17 @@ impl WebSocketState {
     fn is_active(&self) -> bool {
         match *self {
             WebSocketState::Active => true,
+            _ => false,
+        }
+    }
+
+    /// Tell if we should process incoming data. Note that if we send a close frame
+    /// but the remote hasn't confirmed, they might have sent data before they receive our
+    /// close frame, so we should still pass those to client code, hence ClosedByUs is valid.
+    fn can_read(&self) -> bool {
+        match *self {
+            WebSocketState::Active     |
+            WebSocketState::ClosedByUs => true,
             _ => false,
         }
     }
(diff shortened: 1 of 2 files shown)


application-developer-DA:
> One way I can think of that can make this more comprehensible is using an actual state machine, with states and transitions. That can allow keeping actual work separate (like in methods) from the logic of the state machine. It's just one idea. I do think that if this was my code I would like to rewrite it. But you never really know if you will find a better solution before you try, it's just a gut feeling I have when looking at code.

Could be. This particular file which contains all these transitions between different states inside the websocket context also looks dirty to me... I did not have chance to rewrite and retest it though... But it might be, that some changes will be required to make it ready for new tokio.

> The real problem here I think is the lack of tests and documentation. Complex code almost always has bugs, unless it's thoroughly tested, and those tests also make it safe to refactor and experiment with the design. I haven't really looked into autobahn tests, because maybe that covers a lot. Although it didn't seem to cover the issues I filed pull requests for yesterday.

We have unit tests on a lower level to verify the most crucial parts and the the autobahn tests ensure that we comply with the protocol and don't have any issues. At the moment we released the library, our implementation was the most strict implementation we could find among other implementations in Rust. All autobahn tests were green except a couple minor tests which were "non-strict". We also have some fuzzing tests and there were no critical bugs found after some fuzzing (which also was mentioned here: https://www.reddit.com/r/rust/comments/8zpp5f/auditing_popular_crates_how_a_oneline_unsafe_has/).

The issues you found and fixed are fine, but things like "do not process messages after receiving close" are not strict according to RFC (as you've already mentioned). This is probably the reason why such small (yet important for some rare cases) issues have not been spotted before.

najamelan:
I did have a closer look at the websocket crate lately and I must say that it's design is a lot cleaner and simpler. I suggest you review their code when thinking about tungstenite's design. 

The major thing that keeps their design clean is indeed what I mentioned above, they do not automatically respond to ping and close. The user has to do that manually. 

The problem with websocket right now is that they use hyper for the http handshake and they are not compatible with recent versions of hyper. It seems they are working on that. I have no idea how both crates compare in performance.

PR #75, opened 2019-09-12
Don't swallow ping/pong while waiting for close acknowledgement

najamelan:
I didn't notice yesterday, but it also swallows ping/pong frames. Since ping/pong frames can hold user data, it seems more consistent to also let them through.

I defer to the other PR to make them protocol errors, in line with the data frames.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -383,13 +383,15 @@ impl WebSocketContext {
                         OpCtl::Reserved(i) => Err(Error::Protocol(
                             format!("Unknown control frame type {}", i).into(),
                         )),
-                        OpCtl::Ping | OpCtl::Pong if !self.state.is_active() => {
-                            // No ping processing while closing.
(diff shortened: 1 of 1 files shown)


agalakhov:
The RFC says:

>    If an endpoint receives a Ping frame and has not yet sent Pong frame(s) in response to previous Ping frame(s), the endpoint MAY elect to send a Pong frame for only the most recently processed Ping frame.

So even if the sender doesn't swallow ping, the receiver may do so.

najamelan:
Thanks for looking into it! 

Maybe there is some confusion, what this does is the same as before in terms of responding to ping, it is just that if we receive them after we send a CloseFrame, but while waiting for the remote to acknowledge the close handshake, to still return them from `read_message` instead of swallowing them. 

The main point is to make it consistent with the behavior for data frames which was merged yesterday. For pong it might actually matter to client code since ping and pong can hold user data, but realistically it's really about consistency. 

#74 proposes to consider a remote sending anything after having sent a closeframe a protocol error, for stricter compliance with the RFC.

Issue #78, opened 2019-09-14
User issued pong messages should probably be kept separate from replies to pings.


najamelan:
The RFC describes two uses for pong messages:
- reply to ping
- unidirectional heartbeat

Tungstenite automatically replies to pings, and since the RFC defines that the application data should be identical to the ping, so there's not much the user can add to that. For that reason afaict a user should never manually respond to ping.

However the user can send pongs as unidirectional heartbeats. The problem is that right now by storing these on `self.pong`, they can override replies to ping, and potentially have different application data inside, so not constitute a valid response to ping.

It's pretty unwieldy for client code to avoid this problem. If using tungstenite directly it can be avoided by following the reception of a ping by calling `write_pending` until it doesn't return wouldblock to be sure the ping has been answered and only then send a manual pong.

If using through tokio-tungstenite or other code that splits reader and writer tasks, things become complicated, because now you have to communicate from your reader that received the ping that your writer should call `write_pending` before sending the next pong, and through tokio-tungstenite `write_pending` isn't exactly accessible.

I feel it would be better to document that a user should never respond to ping manually and consider user created pong messages like any other message to be queued. It doesn't need to jump the queue. The RFC only says replies to pings should be sent as fast as practical.

That would avoid this problem all together.

application-developer-DA:
Makes sense.

Indeed, previously we "did not notice" this problem as first versions of `tungstenite-rs` did not allow to construct ping and pong messages directly.

mycorrhizas:
@application-developer-DA are there any examples of how to manually construct a pong response? My use case requires that  a pong is sent back with an internal message. From my understanding, this lib automatically returns an empty pong? Can it be overridden?  thanks

modestmicha:
> @application-developer-DA are there any examples of how to manually construct a pong response? My use case requires that a pong is sent back with an internal message. From my understanding, this lib automatically returns an empty pong? Can it be overridden? thanks

To send a custom pong message instead of the default pong message as a reply to a ping you can simply:
```rust
match socket.read_message()? {
    Message::Ping(payload) => {
        socket.write_message(Message::Pong(your_custom_pong_vec_u8))?;
    }
}
```

However, if you wish to keep sending the default pong message and send an additional pong you'd have to:
```rust
match socket.read_message()? {
    Message::Ping(payload) => {
        // Make sure default pong messages are not overwritten.
        socket.write_pending()?;
        // Additional pong here or anywhere else.
[...]

Issue #77, opened 2019-09-14
Review closing the connection


najamelan:
In a similar idea of #76 I would like to have a look at this part of `WebSocketContext::write_pending`:

```rust
// If we're closing and there is nothing to send anymore, we should close the connection.
if self.role == Role::Server && !self.state.can_read() {
    // The underlying TCP connection, in most normal cases, SHOULD be closed
    // first by the server, so that it holds the TIME_WAIT state and not the
    // client (as this would prevent it from re-opening the connection for 2
    // maximum segment lifetimes (2MSL), while there is no corresponding
    // server impact as a TIME_WAIT connection is immediately reopened upon
    // a new SYN with a higher seq number). (RFC 6455)
    self.state = WebSocketState::Terminated;
    Err(Error::ConnectionClosed)
} else {
    Ok(())
}
```
Note that right now the code still does a check on `WebSocketState::ClosedByPeer | WebSocketState::CloseAcknowledged`, but that's exactly what `can_read` covers, namely that we received a close frame from the remote, so I cleaned that up a bit.

This code mixes 2 concerns:
- dealing with closing the underlying connection, which is earlier for server than client. As far as I can tell from the code, we only set state to Terminated when it's safe to drop the underlying connection, and when we do that we return `ConnectionClosed` to indicate that to client code. 
- moving our state forward because now the close handshake is finished. It stands out now that it's kind of strange that we check whether we can read even though we are in `write_pending`

What causes this I think is that there is no state in `WebSocketState` to indicate the close handshake is finished when the remote initiated it. `CloseAcknowlegded` is only used when we initiated. We know that when we received an initial close frame from the remote, we will have cued a reply so this kind of bumps it to terminated, but only for server role.

It doesn't really make sense to have `CloseAcknowlegded` only count when we initiated the close handshake. The websocket protocol mainly distinguishes a state when the handshake is finished, regardless of who initiated it. I think it's just the way it is because given the design of tungstenite, it's kind of hard to do this right. We know when we cue a close response, but we never know when we actually sent it out. It seems not easy to fix this without creating extra overhead on each call to write_pending, similar to how it's done for pong messages for example.

There is a few problems:
[...]

Issue #76, opened 2019-09-13
Review the implementation of WebSocketContext::read_message


najamelan:
Following somewhat from the [discussion here](https://github.com/snapview/tungstenite-rs/pull/72#issuecomment-530927677), but I opened a separate issue, because I really wonder whether this does what it's author intended.

```rust
/// Read a message from the provided stream, if possible.
///
/// This function sends pong and close responses automatically.
/// However, it never blocks on write.
pub fn read_message<Stream>(&mut self, stream: &mut Stream) -> Result<Message>
where
    Stream: Read + Write,
{
    // Do not read from already closed connections.
    self.state.check_active()?;
    loop {
        // Since we may get ping or close, we need to reply to the messages even during read.
        // Thus we call write_pending() but ignore its blocking.
        self.write_pending(stream).no_block()?;
        // If we get here, either write blocks or we have nothing to write.
        // Thus if read blocks, just let it return WouldBlock.
        if let Some(message) = self.read_message_frame(stream)? {
            trace!("Received message {}", message);
            return Ok(message);
        }
    }
}
```
The loop seems like an iffy control structure here. After looking through the methods it calls, if I'm not mistaking, it can only loop if there is a fragmented message being read. However the comment suggests this will deal with responding to ping and close. 

So this will not respond to ping and close directly, but only on the next read. That's inconvenient for client code, especially when using through tokio-tungstenite, which means not being able to call `read_message` directly when a ping or close is returned here.

It also means that `write_pending` is called in a loop when a framed message is being read. That doesn't seem very useful. I would think in the least the call to `write pending` should be before the loop, if we really want to only respond to close/ping from the last read.

Nicer would be to respond to close/ping directly. It's not so clear to me what the best solution is. Something like this comes to mind:
```rust
// This loops so we continue reading frames when a fragmented message is coming in.
// There are 4 conditions that break from this loop:
// - errors
// - ConnectionClosed
// - nothing to read anymore (WouldBlock)
// - a complete message was read
loop {
    if let Some(message) = self.read_message_frame(stream)? {
        trace!("Received message {}", message);
        match message
        {
            // Since we may get ping or close, we need to reply to the messages even during read.
            // Thus we call write_pending() but ignore its blocking.
            Close(..) | Ping(..) => self.write_pending(stream).no_block()?,
            _ => {}
        };
        return Ok(message);
    }
}
```

[...]

PR #79, opened 2019-09-14
WIP: Add some documentation to WebSocket and Error.

najamelan:
So as I have been figuring out from the source code how tungstenite works, I felt it would be silly to just let that disappear when my memory fades, so I wrote documentation, hoping it will also save others time.

This isn't polished:
- `read_message`, `write_pending` and `close` should probably have an Errors section which lists all the possible outcomes
- some of it can hopefully be simplified after other issues are resolved, like: #78. 
- some of the error handling hasn't actually been resolved, like: #74 and the docs will need updating when it does 
- design improvements might make things more convenient, like: #76 which might also change behavior as described here.
- `write_pending` is a user facing method that does not use `check_active`. I haven't really dug into what that really implies. It probably returns `Ok(())` forever for clients, instead of returning `ConnectionClosed`. It should probably be looked into and docs updated.
- I removed the part of doc from `Error::AlreadyClosed` that claimed this would be returned when writing after receiving a Close message, since it was simply not true, but this error kind should probably be re-evaluated after discussion in #77.

I have no intention of polishing this, cause I have my own chores to do on my own libraries, but "Allow edits from maintainers" is on. Please use it.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -25,30 +25,40 @@ pub type Result<T> = result::Result<T, Error>;
 /// Possible WebSocket errors
 #[derive(Debug)]
 pub enum Error {
-    /// WebSocket connection closed normally
+    /// WebSocket connection closed normally. This informs you of the close.
+    /// It's not an error as such and nothing wrong happened.
     ///
-    /// Upon receiving this, the server must drop the WebSocket object as soon as possible
-    /// to close the connection.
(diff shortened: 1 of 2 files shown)


najamelan:
@application-developer-DA Thanks for working through all of my feedback. I know it's long and I know I'm a nitpick, but I hope tungstenite is really a better library for it.
;)

PR #81, opened 2019-09-20
Make use of can_read to clean up some code

najamelan:
This uses the new method to make some code briefer and some conditional simpler.

There are a few other places where `WebSocketState::ClosedByPeer | WebSocketState::CloseAcknowledged` exist, but because of the specific constructs in those places it's not convenient to replace them with `can_read`.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -335,13 +335,8 @@ impl WebSocketContext {
         // If we get to this point, the send queue is empty and the underlying socket is still
         // willing to take more data.
 
-        let closing_state = match self.state {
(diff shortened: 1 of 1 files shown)


najamelan:
> Ok, as I can see the only difference if we compare it to the previous implementation is that here the Terminated is "implicitly" checked as well
Yes, that's true. If coming through `write_message` or `read_message` we already check terminated with a call to `check_active`.

However I when I made this, I missed that we also expose `write_pending` directly to the user which doesn't do this. That means that before, we would just return `Ok(())` where now we will return `ConnectionClosed`. In principle for consistency it might be better to return `AlreadyClosed` in that case.

PR #106, opened 2020-02-01
Allow firefox to connect

macpp:
When connecting, firefox always sends `Connection: keep-alive, Upgrade` header, so connection is always rejected


agalakhov:
That's very interesting, we use Firefox and it always worked for us. How is Firefox configured?

sdroege:
I can confirm this here with Firefox 72.0.2 on Linux, no special configuration.

Would probably make sense to make a new release with this :)

macpp:
I had this problem on linux (ff 74 nightly) and it seems that it also sends `keep-alive` on my windows 10 (ff 69 and 72). No special configuration in both cases.

Issue #107, opened 2020-02-11
Allow establishing of websocket with web optional protocols parameter 


jadedevin13:
cargo run --example server

1. Open web browser dev console.
2. Use ws and not secure wss, otherwise I get invalid token.
3. Run this:

   ```js
   var exampleSocket = new WebSocket("ws://www.example.com/socketserver");
   ```

4. Error:
   ```rust
   thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: 
   HandshakeError::Failure(Protocol("invalid token"))', src/libcore/result.rs:1165:5
   ```

Reference
https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket

rofrol:
Above example does not use optional parameter

Voronar:
@jadedevin13 
I use connection callback (https://github.com/snapview/tungstenite-rs/blob/master/examples/server.rs#L12) as escape hatch:
```rust
req.headers().get("sec-websocket-protocol").map(|sp| {
  let headers = resp.headers_mut();
  headers.append("sec-websocket-protocol", sp.clone());
});
```

application-developer-DA:
Current state: https://github.com/snapview/tungstenite-rs/issues/145#issuecomment-713581499

PR #108, opened 2020-02-11
Allows establishing of websocket with web optional protocols parameter

jadedevin13:
The browser won't autoclose the connection since it will now see the header it requires.

https://github.com/snapview/tungstenite-rs/issues/107


agalakhov:
Sorry, the code you just pushed is not compileable.

application-developer-DA:
There seems to be no activity in this PR, so I'm going to close it. We could re-introduce the changes if we decide to use this approach. Please check https://github.com/snapview/tungstenite-rs/issues/145#issuecomment-713581499 for further details.

PR #117, opened 2020-04-27
connect support timeout

errord:
connect support timeout

### DIFF
--- a/examples/autobahn-client.rs
+++ b/examples/autobahn-client.rs
@@ -6,7 +6,7 @@ use tungstenite::{connect, Error, Message, Result};
 const AGENT: &str = "Tungstenite";
 
 fn get_case_count() -> Result<u32> {
-    let (mut socket, _) = connect(Url::parse("ws://localhost:9001/getCaseCount").unwrap())?;
+    let (mut socket, _) = connect(Url::parse("ws://localhost:9001/getCaseCount").unwrap(), None)?;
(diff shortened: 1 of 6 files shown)


application-developer-DA:
Hm, this is ok, but would break the compatibility with the current version. @agalakhov I need your opinion -- would you like to have it like this or should we instead introduce the function with `timeout` parameter parallel to the existing one, so that we stay backwards-compatible?

application-developer-DA:
@agalakhov I'm still thinking that probably doing a separate function for this instead of modifying our existing one would be a better idea, what do you think? Do we want to have this possibility at all? The `TcpStream` has different functions for that, so I think having a separate function (if at all supporting this) would be a better turn.

agalakhov:
Yes, it should be `connect_with_timeout()` or something like that. The ordinary `connect()` function could be a wrapper with infinite timeout.

application-developer-DA (changes requested):
It would be nice to implement changes from https://github.com/snapview/tungstenite-rs/pull/117#issuecomment-669272772 to be able to merge this PR.

If you think that it does not make sense, feel free to use `tokio-tungstenite` as it allows having such flexibility.

application-developer-DA:
It looks like this PR is obsolete and there was no feedback since October, so I'm going to close it. Feel free to open a new one (rebased to master ;)

PR #119, opened 2020-05-08
Fix ws request subportocol lowercase header

kids300:
[RFC6445 Sec11.3](https://tools.ietf.org/html/rfc6455#section-11.3) specifies several headers like `Sec-WebSocket-Protocol`. While it does not specifies the key should be case insensitive, many server only accepts case sensitive requests. However [http](https://docs.rs/http/0.2.1/http/) case every key to lowercase for convince. [hyper](https://docs.rs/hyper/0.13.5/hyper/) has an issue to solve this problem (https://github.com/hyperium/hyper/issues/1492), but not in http. This is a workaround to make the websocket server happy.

### DIFF
--- a/src/handshake/client.rs
+++ b/src/handshake/client.rs
@@ -138,7 +138,7 @@ fn generate_request(request: Request, key: &str) -> Result<Vec<u8>> {
 
     for (k, v) in request.headers() {
         let mut k = k.as_str();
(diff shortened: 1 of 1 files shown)


application-developer-DA:
@agalakhov , since there were no objections, I'll merge this PR, it does not seem like it can break something, but instead allow more servers to work well with tungstenite.

PR #125, opened 2020-05-21
Change Into to From for CloseCode to u16 conversation

Dushistov:
As stated in standard library documentation:
> One should always prefer implementing From over Into because implementing From automatically
> provides one with an implementation of Into thanks to the blanket implementation in the standard library.

The idea of this implementation to make possible to use `u16::from(close_code)` in closure


agalakhov:
Thank you!

The reason it was `into`, not `from`, is that `from` caused trait implementation conflict in some ancient version of Rust.

Issue #118, opened 2020-05-02
Passing vendored feature to native-tls


TJRoh01:
It is sometimes necessary to statically link OpenSSL when compiling which is achieved by activating the "venodred" feature in the openssl crate. For this reason native-tls, that you use as a dependency, also has this feature that it simply passes to openssl when compiling. 

I would really appreciate it if you could add similar functionality in tungstenite.

application-developer-DA:
Seems to be solved now.

PR #130, opened 2020-06-17
Allow statically linking

Krisa:
Solving https://github.com/snapview/tungstenite-rs/issues/118 by using `features= ["tls-vendored"]` in the toml file


application-developer-DA:
Thanks!

Krisa:
@application-developer-DA thank you for the merge as well. Any chance you push it on `crates.io` as well when you get the chance? I'm referencing this crate from another one and would love to be able to stop managing these dependencies locally :-)

PR #135, opened 2020-08-19
Enable dependabot

RAnders00:
Dependabot will automatically create pull requests suggesting updates to dependencies.


agalakhov:
Applied, thanks.

PR #143, opened 2020-09-20
Some simplifications

phimuemue:
* Avoid unneeded temporary `fin`
* Delegate work in `into_client_request`

### DIFF
--- a/src/protocol/message.rs
+++ b/src/protocol/message.rs
@@ -37,20 +37,16 @@ mod string_collect {
             let mut input: &[u8] = tail.as_ref();
 
             if let Some(mut incomplete) = self.incomplete.take() {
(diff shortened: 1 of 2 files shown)


src/protocol/message.rs
```diff
@@ -37,21 +37,17 @@ mod string_collect {
...
                     }
-                    true
                 } else {
                     input = &[];
-                    false
-                };
-                if !fin {
                     self.incomplete = Some(incomplete)
-                }
+                };
```
hellow554, line 50: Is that `;` necessary?

hellow554, line 50: You could/should move the `;` to the `... = Some...` line

application-developer-DA:
Thanks.

PR #144, opened 2020-09-24
Permessage-deflate extension

SirCipher:
This PR adds support for permessage-deflate for both inbound and outbound messages through the use of WebSocket extensions. 
I've also added this support to the tokio-tungstenite crate in a separate PR.

This is feature-gated through the `deflate` flag and uses the Flate2 crate for inflating and deflating messages.

I've updated the client and server Autobahn results and all unit tests for both pass.


Cargo.toml
```diff
@@ -30,10 +31,24 @@ sha-1 = "0.9"
 url = "2.1.0"
 utf-8 = "0.7.5"
 
+[dependencies.flate2]
+optional = true
+version = "1.0"
+default-features = false
+features = ["zlib"]
```
application-developer-DA, line 38: BTW, do we want to rely on `zlib`? @agalakhov, what do you think?

README.md
```diff
@@ -58,7 +58,7 @@ Features
 Tungstenite provides a complete implementation of the WebSocket specification.
 TLS is supported on all platforms using native-tls.
 
-There is no support for permessage-deflate at the moment. It's planned.
+Permessage-deflate.
```
application-developer-DA, line 61: Perhaps we can remove this line now ;)

src/protocol/mod.rs
```diff
@@ -601,6 +601,13 @@ impl WebSocketContext {
             }
         }
 
+        if frame.header().is_final {
+            frame = match self.config.encoder.on_send_frame(frame) {
+                Ok(frame) => frame,
+                Err(e) => return Err(e.into()),
+            };
```
application-developer-DA, line 608: If I'm not missing something, I think this could be rewritten as:

```rust
frame = self.config.encoder.on_send_frame(frame)?;
```

src/protocol/mod.rs
```diff
@@ -489,49 +528,10 @@ impl WebSocketContext {
...
-                        }
-                        OpData::Reserved(i) => Err(Error::Protocol(
-                            format!("Unknown data frame type {}", i).into(),
-                        )),
-                    }
-                }
+                _ => match self.config.encoder.on_receive_frame(frame) {
+                    Ok(r) => Ok(r),
+                    Err(e) => Err(e.into()),
+                },
```
application-developer-DA, line 534: This seems to be the same as `self.config.encoder.on_receive_frame(frame).map_err(Into::into)`, although i'm not sure that even `map_err()` would be required.

src/protocol/mod.rs
```diff
@@ -426,17 +476,6 @@ impl WebSocketContext {
...
-            // for non-zero values.  If a nonzero value is received and none of
-            // the negotiated extensions defines the meaning of such a nonzero
-            // value, the receiving endpoint MUST _Fail the WebSocket
-            // Connection_.
-            {
-                let hdr = frame.header();
-                if hdr.rsv1 || hdr.rsv2 || hdr.rsv3 {
-                    return Err(Error::Protocol("Reserved bits are non-zero".into()));
-                }
-            }
```
application-developer-DA, line 439: Does it have any adverse effects that we postpone this check?

src/protocol/mod.rs
```diff
@@ -489,49 +528,10 @@ impl WebSocketContext {
...
-                                self.incomplete = Some(msg);
-                                Ok(None)
-                            }
-                        }
-                        OpData::Reserved(i) => Err(Error::Protocol(
-                            format!("Unknown data frame type {}", i).into(),
-                        )),
-                    }
-                }
+                _ => match self.config.encoder.on_receive_frame(frame) {
```
application-developer-DA, line 531: Why `_` instead of `OpCode::Control`?

src/handshake/client.rs
```diff
@@ -90,31 +103,45 @@ impl<S: Read + Write> HandshakeRole for ClientHandshake<S> {
...
+    key: &str,
+    config: &mut Option<WebSocketConfig<Ext>>,
+) -> Result<Vec<u8>>
+where
+    Ext: WebSocketExtension,
+{
+    let request = match config {
+        Some(ref mut config) => config.encoder.on_make_request(request),
+        None => request,
+    };
```
application-developer-DA, line 129: Could it be rewritten like this?

`let request = config.map(|config| config.encoder.on_make_request(request)).unwrap_or(request);`

It would probably look cleaner like this.

src/handshake/client.rs
```diff
@@ -90,31 +103,45 @@ impl<S: Read + Write> HandshakeRole for ClientHandshake<S> {
                 result,
                 tail,
             } => {
-                self.verify_data.verify_response(&result)?;
+                let mut config = self.config.take().unwrap();
```
application-developer-DA, line 106: Oh, that looks like a bold move :) Are we sure that it's always `Some`? Because if we do, then probably `config` must not be an `Option`.

src/handshake/client.rs
```diff
@@ -205,12 +239,18 @@ impl VerifyData {
...
         // field and this header field indicates the use of an extension
         // that was not present in the client's handshake (the server has
         // indicated an extension not requested by the client), the client
         // MUST _Fail the WebSocket Connection_. (RFC 6455)
-        // TODO
+
+        if let Some(config) = config {
+            if let Err(e) = config.encoder.on_response(response) {
+                return Err(e.into());
+            }
```
application-developer-DA, line 252: I think it can be rewritten as:

```rust
config.map(|conf| conf.encoder.on_response(response)).transpose()?;
```

src/handshake/client.rs
```diff
@@ -52,15 +60,15 @@ impl<S: Read + Write> ClientHandshake<S> {
...
+            let req = generate_request(request, &key, &mut config)?;
             HandshakeMachine::start_write(stream, req)
         };
 
         let client = {
             let accept_key = convert_key(key.as_ref()).unwrap();
             ClientHandshake {
                 verify_data: VerifyData { accept_key },
-                config,
+                config: Some(config),
```
application-developer-DA, line 71: It looks a big counter-intuitive to me: why do we make it `Some(...)` if it's stored as an `Option`? Something is wrong with it. We either should not pass `Some` or not store it as `Option`? Or am I missing something?

src/handshake/server.rs
```diff
@@ -231,16 +252,23 @@ impl<S: Read + Write, C: Callback> HandshakeRole for ServerHandshake<S, C> {
...
                 if !tail.is_empty() {
                     return Err(Error::Protocol("Junk after client request".into()));
                 }
 
-                let response = create_response(&result)?;
+                let mut response = create_response(&request)?;
+
+                if let Some(ref mut config) = self.config.as_mut().unwrap() {
+                    if let Err(e) = config.encoder.on_receive_request(&request, &mut response) {
+                        return Err(e.into());
```
application-developer-DA, line 266: See error handling mentioned below.

src/protocol/mod.rs
```diff
@@ -229,38 +270,47 @@ pub struct WebSocketContext {
...
 
-impl WebSocketContext {
+impl<Ext> WebSocketContext<Ext>
+where
+    Ext: WebSocketExtension,
+{
     /// Create a WebSocket context that manages a post-handshake stream.
-    pub fn new(role: Role, config: Option<WebSocketConfig>) -> Self {
+    pub fn new(role: Role, config: Option<WebSocketConfig<Ext>>) -> Self {
+        let config = config.unwrap_or_else(Default::default);
```
application-developer-DA, line 282: Out of curiosity: why did you decide to move out the config creation line?

src/extensions/uncompressed.rs
```diff
@@ -0,0 +1,104 @@
...
+                    } else {
+                        self.incomplete = Some(msg);
+                        Ok(None)
+                    }
+                }
+                Data::Reserved(i) => Err(Error::Protocol(
+                    format!("Unknown data frame type {}", i).into(),
+                )),
+            },
+            _ => unreachable!(),
```
application-developer-DA, line 101: It's a bit dangerous to panic here as we don't really know (from the compiler's point of view there is no guarantee for that) that it's really `unreachable!()`. `unreachable!()` statements are ok in some cases, when the place it's used is close to the code that ensures the invariant. But here we don't really have any guarantee that we never get here.

src/extensions/deflate.rs
```diff
@@ -0,0 +1,922 @@
+//! Permessage-deflate extension
```
application-developer-DA, line 1: Did you write this file manually? Or is it a modified copy-paste from `deflate`? I'm asking, because it's quite a huge file and I and @agalakhov will have to validate it carefully as it contains `loop`s, `unwrap`s and some copy-pasting.

application-developer-DA (changes requested):
First of all: thanks for the PR and the changes that you made!

I left some comments, which I think must be addressed before we can merge them. I also have not yet checked `deflate.rs` implementation.

Also I have one general question: do we really need an additional type parameter in all places? Maybe we can avoid it? Having a generic trait for extensions is fine, but can't we use many extensions and is a `Config` a right place to choose the extension? I'm not sure at the moment. It seems like the current implementation assumes that `encoder` is just a concrete implementation of permessage-deflate, rather than a generic extension or list of extensions.

@agalakhov could you please check the comments that I left to this PR and check if I missed something? This is quite a huge PR and I would check it carefully before we can merge it. But I think there is some work that has to be done.

SirCipher:
Hi @application-developer-DA, thank you for the review. Regarding the additional type parameter, the thought behind this was that it allowed for users to implement their own extensions and provide them. An alternative approach to this would be to have permessage-deflate configurable in the `WebSocketConfig` structure and then select the appropriate encoder when building the `WebSocket`. If this is something that you would prefer, then I'm happy to make the change, and then that would resolve a number of the comments that you have made. 

Regarding your error handling/into comments, the `?` operator doesn't work with the trait's associated type for the error - which is why I matched on the error and then did `Into::into`, though this could be tidied up with `map_err`.

src/protocol/mod.rs
SirCipher: The blocks that you've mentioned this on could be tidied up using `map_err`, but they don't work with `?` and the trait's associated type.

src/protocol/mod.rs
SirCipher: If the deflate extension is disabled, then it delegates the work to the uncompressed extension and does not mutate the frame in any way.

src/handshake/client.rs
SirCipher: Upon the handshake finishing, the contents of the option are removed and the `WebSocket` instance is created. With some additional work, this could possibly be tidied up. Or, if the functionality to provide custom implementations of extensions is removed, then this will be removed.

src/extensions/uncompressed.rs
SirCipher: It is rather tightly coupled with the caller in that regard. Perhaps the frame could be deconstructed into the required parts to remove the `unreachable` call.

src/extensions/deflate.rs
SirCipher: It's an adaptation of WS-RS's deflate extension to use ZLib instead of libc, and to work with a growable fragment buffer.

application-developer-DA:
@SirCipher , yes, I think having it inside a config makes sense. It seems like it would simplify many things. What do you think?

src/extensions/uncompressed.rs
application-developer-DA: Yeah, that's how I imagined it.

src/extensions/deflate.rs
application-developer-DA: Got it. I think I or @agalakhov will have to take a close look at it. At the first glance, it seems like it contains some copy-paste that we would like to avoid :)

novacrazy:
What's the status of this PR?

application-developer-DA:
@novacrazy the PR does not pass all of our CI/CD checks and has some conflicts, thus it can't be merged. I'm going to look at the content one more time to see if the rest is fine.

application-developer-DA (changes requested):
Thanks for doing that great progress, but unfortunately there are several reasons why we can't merge it right now:

1. The PR does not pass our current CI/CD pipeline.
2. It looks like part of this PR should be done as a separate library, i.e. I think that the whole deflate/compression/uncompression logic here should not be part of `tungstenite-rs` as it could be generalized and does not have anything to do with WebSockets. This would make the implementation much more simple and concise. I'm afraid that if we merge it "as is" (even if (1) is fixed), we would introduce some technical debt to the library which we would like to avoid if possible, unless I'm missing on something.

SirCipher:
Due to time constraints on my end and the additional changes required to make the requested changes, I won't be able to finish this PR in the near future.

PR #148, opened 2020-10-01
Add facilities to allow clients to follow HTTP 3xx redirects

Redrield:
Adds Error::Redirection, which is a special case of Error::Http that
extracts the redirection target from the response headers, and stores it
in the error object. Client implementations that build upon tungstenite
can use this to implement redirecting. The connect() function defined in this crate will automatically follow
redirecting responses. Its error handler causes it to recurse when a value of Error::Redirection is returned.

A catch-all solution for redirects is not possible due to the
abstraction transforming socket types to Read + Write, implementations
that use the client_* methods need to handle redirections themselves.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -91,6 +91,12 @@ pub fn connect_with_config<Req: IntoClientRequest>(
     config: Option<WebSocketConfig>,
 ) -> Result<(WebSocket<AutoStream>, Response)> {
     let request: Request = request.into_client_request()?;
+    // Copy all the fields from the initial reqeust **except** the URI. This will be used in the event of a redirection code
+    // Have to manually clone Method because there is one field that contains a Box,
+    // but in the case of normal request methods it is Copy
+    let request2 = Request::builder()
+        .method(request.method().clone())
+        .version(request.version());
     let uri = request.uri();
     let mode = uri_mode(uri)?;
     let host = request
@@ -104,10 +110,20 @@ pub fn connect_with_config<Req: IntoClientRequest>(
     let addrs = (host, port).to_socket_addrs()?;
     let mut stream = connect_to_some(addrs.as_slice(), &request.uri(), mode)?;
     NoDelay::set_nodelay(&mut stream, true)?;
-    client_with_config(request, stream, config).map_err(|e| match e {
+    match client_with_config(request, stream, config).map_err(|e| match e {
         HandshakeError::Failure(f) => f,
         HandshakeError::Interrupted(_) => panic!("Bug: blocking handshake not blocked"),
-    })
+    }) {
+        Ok(r) => Ok(r),
+        Err(e) => match e {
+            Error::Redirection(uri) =>  {
+                debug!("Redirecting to {}", uri);
+                let request = request2.uri(uri).body(()).unwrap();
+                connect_with_config(request, config)
+            }
+            _ => Err(e),
+        }
+    }
 }
 
 /// Connect to the given WebSocket in blocking mode.
--- a/src/client.rs
+++ b/src/client.rs
@@ -4,7 +4,7 @@ use std::io::{Read, Write};
 use std::net::{SocketAddr, TcpStream, ToSocketAddrs};
 use std::result::Result as StdResult;
 
-use http::Uri;
+use http::{Uri, request::Parts};
 use log::*;
 
 use url::Url;
@@ -89,10 +89,12 @@ use crate::stream::{Mode, NoDelay};
 pub fn connect_with_config<Req: IntoClientRequest>(
     request: Req,
     config: Option<WebSocketConfig>,
+    max_redirects: u8,
 ) -> Result<(WebSocket<AutoStream>, Response)> {
-    let mut request: Request = request.into_client_request()?;
 
-    fn inner(request: Request, config: Option<WebSocketConfig>) -> Result<(WebSocket<AutoStream>, Response)> {
+    fn try_client_handshake(request: Request, config: Option<WebSocketConfig>)
+        -> Result<(WebSocket<AutoStream>, Response)>
+    {
         let uri = request.uri();
         let mode = uri_mode(uri)?;
         let host = request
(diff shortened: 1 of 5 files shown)


src/handshake/client.rs
```diff
@@ -160,7 +160,12 @@ impl VerifyData {
         // 1. If the status code received from the server is not 101, the
         // client handles the response per HTTP [RFC2616] procedures. (RFC 6455)
         if response.status() != StatusCode::SWITCHING_PROTOCOLS {
-            return Err(Error::Http(response.status()));
+            if response.status().is_redirection() {
+                let value = response.headers().get("Location").unwrap();
```
application-developer-DA, line 164: While `Location` must be set in normal cases, we can't rely on that (someone could have implemented a bad web server that does not behave properly => we must not panic in such cases). I think the safer approach would be to check the existence of header.

Also we introduce a new variant on the `Error` for `Redirect`ing, but if I'm not missing on something, then it won't appear in the public API, i.e. no-one will ever be able to get this error as it's the one used only inside our handshake module which might be confusing for other people (they don't know if redirects are forwarded or not). Maybe we should still return an `Http` error like we did before, but instead of containing the `StatusCode` use the `Response`?

This would allow us to handle redirects at one place (if any).

src/client.rs
```diff
@@ -104,10 +110,20 @@ pub fn connect_with_config<Req: IntoClientRequest>(
...
         HandshakeError::Failure(f) => f,
         HandshakeError::Interrupted(_) => panic!("Bug: blocking handshake not blocked"),
-    })
+    }) {
+        Ok(r) => Ok(r),
+        Err(e) => match e {
+            Error::Redirection(uri) =>  {
+                debug!("Redirecting to {}", uri);
+                let request = request2.uri(uri).body(()).unwrap();
+                connect_with_config(request, config)
```
application-developer-DA, line 122: We need to find a way to limit the amount of redirections to not end up with an infinite loop. Perhaps an additional parameter in the config would do? (if any)

src/client.rs
```diff
@@ -91,6 +91,12 @@ pub fn connect_with_config<Req: IntoClientRequest>(
     config: Option<WebSocketConfig>,
 ) -> Result<(WebSocket<AutoStream>, Response)> {
     let request: Request = request.into_client_request()?;
+    // Copy all the fields from the initial reqeust **except** the URI. This will be used in the event of a redirection code
+    // Have to manually clone Method because there is one field that contains a Box,
+    // but in the case of normal request methods it is Copy
+    let request2 = Request::builder()
+        .method(request.method().clone())
+        .version(request.version());
```
application-developer-DA, line 99: Alternatively we could make a requirement that `IntoClientRequest` also implements `Clone`, then we could just clone it and change the URI which probably would look a bit more elegant.

src/client.rs
```diff
@@ -104,10 +110,20 @@ pub fn connect_with_config<Req: IntoClientRequest>(
     let addrs = (host, port).to_socket_addrs()?;
     let mut stream = connect_to_some(addrs.as_slice(), &request.uri(), mode)?;
     NoDelay::set_nodelay(&mut stream, true)?;
-    client_with_config(request, stream, config).map_err(|e| match e {
+    match client_with_config(request, stream, config).map_err(|e| match e {
```
application-developer-DA, line 113: I think, instead of matching the error one more time after handling it, we could just use `or_else`, it would look nicer.

application-developer-DA (changes requested):
Thanks. The idea looks good, but I think we need some changes to make it robust.

src/client.rs
Redrield: Sure thing, I'll add a limit in the next commit, what should the sensible default for it be?

src/handshake/client.rs
Redrield: I'll fix the header checks in the next commit

If you prefer I can rebase when #150 is merged and build on that, or the changes there can be subsumed into this PR, makes no difference to me.

application-developer-DA: The changes in #150 are minimal, I think we can do them right here.

src/client.rs
application-developer-DA: Maybe something like `max_redirect_count: Option<usize>`, where `None` would mean "not follow redirects", but I'm open to ideas :)

src/client.rs
application-developer-DA: Ah, got it. I just thought that what basically we want to have is to have the same request, but with a different URI, thus the semantics of `request.clone().set_uri(new_uri)`.

Redrield: Request isn't Clone so we'd still have to formulate it like that anyways

application-developer-DA (changes requested):

src/protocol/mod.rs
```diff
@@ -51,6 +54,7 @@ impl Default for WebSocketConfig {
             max_send_queue: None,
             max_message_size: Some(64 << 20),
             max_frame_size: Some(16 << 20),
+            max_redirects: Some(3)
```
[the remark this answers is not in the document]
Redrield: Unless tokio-tungstenite uses the connect functions instead of the client functions to build its API the redirects specified here won't make a difference, since it'd need to implement its own redirect code where the config could be handled.

src/error.rs
```diff
@@ -63,8 +64,12 @@ pub enum Error {
     Utf8,
     /// Invalid URL.
     Url(Cow<'static, str>),
+    /// HTTP error (status only).
+    HttpStatus(StatusCode),
     /// HTTP error.
-    Http(http::StatusCode),
+    Http(Response<()>),
+    /// No Location header in 3xx response
+    NoLocation,
```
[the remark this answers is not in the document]
Redrield: I can scrap the NoLocation error type and just return Error::Http with the given response if Location isn't present but I thought that would be obtuse to debug. If you'd prefer I'll do that

application-developer-DA:
Hey @Redrield , I've just pushed changed to your branch, summary:

1. I refactored the error handling logic, so that we use one error for HTTP errors, not 3 different ones. This should also simplify the debugging of server in case it returns an error.
2. After some thinking, I also refactored the redirect logic a bit, although it still looks a bit ugly.
3. Rebased to master.

I have not checked if it properly works, but looks correct at the first side. What do you think?

Redrield:
Just tested it with the examples, plus a server in the middle returning 302 Found and it does work. The only change that I'd make is having the default number of redirections before abort be at least 1, rather than 0.

application-developer-DA:
@Redrield , alright. I just used `0`, so that the "default behavior" of `connect` is not changed. But then I realized that devs who use `connect()` are likely to be interested in redirects being followed as a default behavior, thus they're calling the high-level `connect` function instead of calling other functions that give a more fine-grained control. So I changed it to `3` as you originally suggested in the previous revisions of this PR. Hope it makes sense.

PR #149, opened 2020-10-07
Update base64 dependency to 0.13

sdroege:


najamelan:
Just be careful, if it's a public dep it probably requires a version bump.

sdroege:
It's not a public dependency whose types are exported from tungstenite, purely internal to do some base64 encoding/decoding.

application-developer-DA:
Thanks @sdroege , but I think dependabot has just updated it :)

(I went through my e-mails and the PR from bot was before yours, so I merged it)

Issue #46, opened 2018-07-23
Handling of HTTP-redirects (follow HTTP 3xx redirects)


shmutalov:
How to handle HTTP redirects? I am trying to connect to gitter API with no success:
https://github.com/shmutalov/gitter-rs/blob/develop-streaming/faye/src/lib.rs#L71

bsergean:
This was filed here too -> #139

Redrield:
If this is something that the maintainers would be fine with making a special case for, I can implement it

application-developer-DA:
I'm a bit indifferent :) But if it does not bring any adverse side effects, then why not?

@agalakhov, what do you think?

application-developer-DA:
Fixed by #148

PR #150, opened 2020-10-14
GH-46 Return the response in case of non-101 response from server

inikulin:
Following redirects inside the lib (https://github.com/snapview/tungstenite-rs/pull/148)
has few flows: there is no redirect loop prevention in this case, in case of using the lib in proxy it's impossible to return the upstream response to browser, etc.

With this change response is propagated to the lib's user, so it can decide what to do with it
in case of redirects: either send it to browser for it to follow redirects or implement redirect following on their side.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -65,6 +65,8 @@ pub enum Error {
     Url(Cow<'static, str>),
     /// HTTP error.
     Http(http::StatusCode),
+    /// HTTP response error.
+    HttpResponse(http::Response<()>),
     /// HTTP format error.
     HttpFormat(http::Error),
 }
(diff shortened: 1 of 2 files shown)


application-developer-DA (approved):
~Looks good to me.~ Discussions needed.

@agalakhov +?

application-developer-DA (commented):
Thanks, this is a good point, I'm actually currently reviewing the https://github.com/snapview/tungstenite-rs/pull/148/files , perhaps the issues handled by this PR will be done within https://github.com/snapview/tungstenite-rs/pull/148/files as well, let's see if they will be required.

inikulin:
@application-developer-DA with that plan, will it be possible to opt-out from built-in redirect following logic?

application-developer-DA:
@inikulin , sure! That was one of the reasons I mentioned that we need to configure the `max_redirects: Option<usize>`, so that there is a possibility to tell `tungstenite`: "Please, do not do automatic redirects".

application-developer-DA:
Superseeded by https://github.com/snapview/tungstenite-rs/pull/148

PR #151, opened 2020-10-14
Respect existing "Sec-WebSocket-Key" in client requests

patr0nus:
This PR allows specifying a custom `Sec-WebSocket-Key` in client request's header. This can be useful if

- The server has some extra requirement for `Sec-WebSocket-Key`, other than just a random base64 value, or
- For some security season the `rand::random()` is not good enough for generating `Sec-WebSocket-Key`.

### DIFF
--- a/src/handshake/client.rs
+++ b/src/handshake/client.rs
@@ -31,7 +31,7 @@ impl<S: Read + Write> ClientHandshake<S> {
     /// Initiate a client handshake.
     pub fn start(
         stream: S,
-        request: Request,
+        mut request: Request,
         config: Option<WebSocketConfig>,
     ) -> Result<MidHandshake<Self>> {
         if request.method() != http::Method::GET {
@@ -49,7 +49,12 @@ impl<S: Read + Write> ClientHandshake<S> {
         // Check the URI scheme: only ws or wss are supported
         let _ = crate::client::uri_mode(request.uri())?;
 
-        let key = generate_key();
+        let key = if let Some(key) = request.headers_mut().remove("Sec-WebSocket-Key") {
+            key.to_str()?.to_owned()
+        }
+        else {
+            generate_key()
+        };
 
         let machine = {
             let req = generate_request(request, &key)?;


src/handshake/client.rs
```diff
@@ -49,7 +49,12 @@ impl<S: Read + Write> ClientHandshake<S> {
         // Check the URI scheme: only ws or wss are supported
         let _ = crate::client::uri_mode(request.uri())?;
 
-        let key = generate_key();
+        let key = if let Some(key) = request.headers_mut().remove("Sec-WebSocket-Key") {
+            key.to_str()?.to_owned()
+        }
+        else {
+            generate_key()
+        };
```
application-developer-DA, line 57: This could be rewritten as `request.headers_mut().remove("Sec-WebSocket-Key").map(|key| ...).unwrap_or_else(|| generate_key())`.

application-developer-DA:
I'm not sure about it to be honest, because according to the RFC:

```
   The request MUST include a header field with the name
        |Sec-WebSocket-Key|.  The value of this header field MUST be a
        nonce consisting of a randomly selected 16-byte value that has
        been base64-encoded (see Section 4 of [RFC4648]).  The nonce
        MUST be selected randomly for each connection.
```

application-developer-DA:
If there are no objections, I think it makes sense to close this PR. FYI @agalakhov 

@patr0nus sorry for that, but we can't implement features that are contrary to the RFC. Let me know if I'm missing on something.

patr0nus:
Hi @application-developer-DA,

I know It's been a long time but I want to discuss a niche (but valid IMO) case of customizing  `Sec-WebSocket-Key`.

I am implementing a proxy server that understands websocket traffic. To do this I use 
tungstenite to accept ws connections from clients and establish ws connections to servers on behalf of clients.

In the process of establishing the ws connection to the server, **I want to send `Sec-WebSocket-Key` using the exact value that the client sends to me.** I believe this is more standard-conforming than generating a new one because `Sec-WebSocket-Key` is a ["end-to-end" header](https://datatracker.ietf.org/doc/html/rfc2616#section-13.5.1). According to the HTTP/1.1 RFC:

```
 - End-to-end headers, which are transmitted to the ultimate recipient of a request or response.

......

   All other headers defined by HTTP/1.1 are end-to-end headers.

   Other hop-by-hop headers MUST be listed in a Connection header,
   section 14.10 to be introduced into HTTP/1.1 (or later).
```

Would you consider reopening this PR?

application-developer-DA:
@patr0nus, I believe it's not required since the release of the version `0.17.0` [where we re-worked](https://github.com/snapview/tungstenite-rs/blob/master/CHANGELOG.md#0170) the behavior of the client request generation 🙂 The change that we introduced in `0.17.0` is even more generic meaning that now you are free to pass/modify **any headers** if you want (**or** rely on `tungstenite` request generation if you want to preserve the behavior that we had in the previous versions). So we now give even more control and choice to the user to specify the behavior that they need while still allowing them to use the "helpers" from the `tungstenite` if necessary.

PR #154, opened 2020-10-22
matches!: use macros; remove unused imports

Horki:
@agalakhov please test


application-developer-DA (approved):
Thanks, that looks much nicer now!

PR #158, opened 2020-11-24
Add a custom response builder

Austaras:
I'd like to build my reponse so I can feed it into my own http pipeline, that's impossible without having the ability of getting a `Reponse<T>`

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -119,10 +119,8 @@ pub fn connect_with_config<Req: IntoClientRequest>(
     }
 
     fn create_request(parts: &Parts, uri: &Uri) -> Request {
-        let mut builder = Request::builder()
-            .uri(uri.clone())
-            .method(parts.method.clone())
-            .version(parts.version);
+        let mut builder =
+            Request::builder().uri(uri.clone()).method(parts.method.clone()).version(parts.version);
         *builder.headers_mut().expect("Failed to create `Request`") = parts.headers.clone();
         builder.body(()).expect("Failed to create `Request`")
     }
--- a/src/handshake/server.rs
+++ b/src/handshake/server.rs
@@ -81,18 +81,12 @@ fn create_parts<T>(request: &HttpRequest<T>) -> Result<Builder> {
 
 /// Create a response for the request.
 pub fn create_response(request: &Request) -> Result<Response> {
-    match create_parts(&request) {
-        Ok(builder) => Ok(builder.body(())?),
-        Err(e) => Err(e)
-    }
+    Ok(create_parts(&request)?.body(())?)
 }
 
 /// Create a response for the request, with a custom body builder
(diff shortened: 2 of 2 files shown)


application-developer-DA (changes requested):
I'm not sure what are you trying to accomplish with that. If you want to define answer that would contain an additional body, you could call `accept_hdr` and pass your own callback to complete headers. Adding some additional body at this place would probably work better than creation of a custom `create_response_t()`.

The submitted code also have a rather non-intuitive name `create_response_t` with `empty` parameter, the implementation of which could have been written like this:

```
Ok(create_parts(&request)?.body(generate_body())?)
```

Though I'm still not sure that it would be required, because it seems like there are other ways to achieve this.

Austaras:
I have a custom http server built upon `hyper` and it needs a `Response<Body>` where `Body` come from `hyper`. However either `accept_hdr` or `create_response` only gives back a `Response<T>`, I could make it into parts and build response again from parts again, but I don't think that's the optimal solution.

I should definitely come up with better names though

application-developer-DA:
Ok, so you have a `hyper` server and at some point you want to upgrade it to the `WebSocket` once you see that the client is trying to upgrade the connection, is my assumption correct?

The correct way would be to make `Response` and `Request` more generic. I'm not sure though how changing `create_response()` would help you as it will only generate a single response that you could forward to the client instead of establishing a `WebSocket` connection. It looks like in your case you would need to implement a state machine in your server using the `ServerHandshake` (after generalizing the `create_response()` and other places to accept some generic `BodyType` instead of `()`).

While this is possible and not really complex to write, I think it is a bit cumbersome to write the state machine from scratch, so I may suggest you to take a look at `warp` which actually implemented this state machine for you and relies on `tokio-tungstenite` for the WebSocket support. That would save you a ton of time in a long run, especially if you're writing some sort of a performant and future-proof server. I'm not sure using bare `tungstenite-rs` at this point makes much of a sense.

Austaras:
For sure I can just use `warp` but that's not fun :)
And yes I'm copying what `warp` is doing, and it's not that bad, which is basically `body.on_upgrade` and then `WebSocket::from_raw_socket`. It uses `headers_ext::SecWebsocketAccept` to build its response, so I believe it would be better if I could reuse your code to achieve the same goal.
Thanks for the detailed response anyway

Austaras:
Aforementioned warp code [here](https://github.com/seanmonstar/warp/blob/29476809294f3b72f7be8d31bdd5e449d433b997/src/filters/ws.rs#L134), and since we are at it, should `tokio-tungstenite` re export `create_response`?

application-developer-DA:
If it's going to be used from other libraries, then yes, though I think it's a rare case to do so, because if library uses `create_response()`, then probably it also does some logic with `MidHandShake`, but since `MidHandShake` is public, the `create_response()` as well. We could re-export it though.

Austaras:
@application-developer-DA  hi, do you have any other suggestions for this pr? If not, may you get this merged? It may be a rare case in the wild, yet it's quite common for web frameworks

PR #159, opened 2020-11-26
WebSocketConfig extended to allow accepting unmasked client frames

adambezecny:
This change will make possible for websocket server to accept unmasked frames. This is needed since some legacy commercially used solutions are unfortunately violating  RFC 6455 and sending (acting as websocket clients)  unmasked client frames. Example is **_Asterisk AudioFork_** (https://github.com/nadirhamid/asterisk-audiofork) used in asterisk based telephony solutions. 

Also other widely used websocket implementations (e.g. **_Node.JS WS_** library - https://www.npmjs.com/package/ws) allow to create WS servers accepting unmasked client frames.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -457,8 +457,7 @@ impl WebSocketContext {
                         // The server MUST close the connection upon receiving a
                         // frame that is not masked. (RFC 6455)
                         // The only exception here is if the user explicitly accepts given
-                        // stream (by tungstenite::server::accept_with_config or tungstenite::server::accept_hdr_with_config)
-                        // with WebSocketConfig.accept_unmasked_frames set to true
+                        // stream by setting WebSocketConfig.accept_unmasked_frames to true
                         if !self.config.accept_unmasked_frames {
                             return Err(Error::Protocol(
                                 "Received an unmasked frame from client".into(),
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -54,7 +54,7 @@ pub struct WebSocketConfig {
     /// Even though this behaviour is not in compliance with RFC 6455 (which requires the server
     /// to close the connection when unmasked frame from client is received) it might be handy in some cases
     /// as there are existing applications sending unmasked client frames.
-    pub server_allow_unmasked: Option<bool>,
+    pub server_allow_unmasked: bool,
 }
 
 impl Default for WebSocketConfig {
@@ -63,7 +63,7 @@ impl Default for WebSocketConfig {
             max_send_queue: None,
             max_message_size: Some(64 << 20),
             max_frame_size: Some(16 << 20),
-            server_allow_unmasked: None,
+            server_allow_unmasked: false,
         }
     }
 }
@@ -457,16 +457,8 @@ impl WebSocketContext {
                         // frame that is not masked. (RFC 6455)
                         // The only exception here is if the user explicitly accepts given
                         // stream (by tungstenite::server::accept_with_config or tungstenite::server::accept_hdr_with_config)
-                        // with WebSocketConfig.server_allow_unmasked set to Some(true)
-                        if let Some(server_allow_unmasked_val) =
-                            self.get_config().server_allow_unmasked
-                        {
-                            if server_allow_unmasked_val == false {
-                                return Err(Error::Protocol(
-                                    "Received an unmasked frame from client".into(),
-                                ));
-                            }
-                        } else {
+                        // with WebSocketConfig.server_allow_unmasked set to true
+                        if self.get_config().server_allow_unmasked == false {
                             return Err(Error::Protocol(
                                 "Received an unmasked frame from client".into(),
                             ));
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -50,6 +50,11 @@ pub struct WebSocketConfig {
     /// be reasonably big for all normal use-cases but small enough to prevent memory eating
     /// by a malicious user.
     pub max_frame_size: Option<usize>,
+    /// If set to true it will allow the websocket server to accept unmasked frames from client.
+    /// Even though this behaviour is not in compliance with RFC 6455 (which requires the server
+    /// to close the connection when unmasked frame from client is received) it might be handy in some cases
+    /// as there are existing applications sending unmasked client frames.
+    pub server_allow_unmasked: Option<bool>,
 }
 
 impl Default for WebSocketConfig {
@@ -58,6 +63,7 @@ impl Default for WebSocketConfig {
             max_send_queue: None,
             max_message_size: Some(64 << 20),
             max_frame_size: Some(16 << 20),
+            server_allow_unmasked: None,
         }
     }
 }
@@ -449,9 +455,22 @@ impl WebSocketContext {
                     } else {
                         // The server MUST close the connection upon receiving a
                         // frame that is not masked. (RFC 6455)
-                        return Err(Error::Protocol(
-                            "Received an unmasked frame from client".into(),
-                        ));
+                        // The only exception here is if the user explicitly accepts given
+                        // stream (by tungstenite::server::accept_with_config or tungstenite::server::accept_hdr_with_config)
+                        // with WebSocketConfig.server_allow_unmasked set to Some(true)
+                        if let Some(server_allow_unmasked_val) =
+                            self.get_config().server_allow_unmasked
+                        {
+                            if server_allow_unmasked_val == false {
+                                return Err(Error::Protocol(
+                                    "Received an unmasked frame from client".into(),
+                                ));
+                            }
+                        } else {
+                            return Err(Error::Protocol(
+                                "Received an unmasked frame from client".into(),
+                            ));
+                        }
                     }
                 }
                 Role::Client => {
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -453,16 +453,14 @@ impl WebSocketContext {
                         // A server MUST remove masking for data frames received from a client
                         // as described in Section 5.3. (RFC 6455)
                         frame.apply_mask()
-                    } else {
+                    } else if !self.config.accept_unmasked_frames {
                         // The server MUST close the connection upon receiving a
                         // frame that is not masked. (RFC 6455)
                         // The only exception here is if the user explicitly accepts given
                         // stream by setting WebSocketConfig.accept_unmasked_frames to true
-                        if !self.config.accept_unmasked_frames {
-                            return Err(Error::Protocol(
-                                "Received an unmasked frame from client".into(),
-                            ));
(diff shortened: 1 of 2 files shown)


src/protocol/mod.rs
```diff
@@ -50,6 +50,11 @@ pub struct WebSocketConfig {
     /// be reasonably big for all normal use-cases but small enough to prevent memory eating
     /// by a malicious user.
     pub max_frame_size: Option<usize>,
+    /// If set to true it will allow the websocket server to accept unmasked frames from client.
+    /// Even though this behaviour is not in compliance with RFC 6455 (which requires the server
+    /// to close the connection when unmasked frame from client is received) it might be handy in some cases
+    /// as there are existing applications sending unmasked client frames.
+    pub server_allow_unmasked: Option<bool>,
```
strohel, line 57: I seems to me that this doesn't have to be an `Option` - the default should be clear from the `Default` implementation. While all other fields are options, they seem to be so to represent "unlimited".

src/protocol/mod.rs
```diff
@@ -449,9 +455,22 @@ impl WebSocketContext {
...
-                        return Err(Error::Protocol(
-                            "Received an unmasked frame from client".into(),
-                        ));
+                        // The only exception here is if the user explicitly accepts given
+                        // stream (by tungstenite::server::accept_with_config or tungstenite::server::accept_hdr_with_config)
+                        // with WebSocketConfig.server_allow_unmasked set to Some(true)
+                        if let Some(server_allow_unmasked_val) =
+                            self.get_config().server_allow_unmasked
+                        {
+                            if server_allow_unmasked_val == false {
```
strohel, line 464: Just some Rust fu: even when server_allow_unmasked is `Option<bool>`, this wouldn't have to be 2 nested ifs - mere `if self.get_config().server_allow_unmasked == Some(true)` should have sufficed.

(this comment is void if we change server_allow_unmasked to plain bool)

strohel (commented):
Thanks for your very first GiHub contribution!

It looks reasonable, I have one minor suggestion. @application-developer-DA please check whether you agree, I'm new to this codebase.

src/protocol/mod.rs
adambezecny: I wasn't sure about this as well and rather wrapped it into Option to keep it consistent. Once you confirm I will make it into simple bool.

src/protocol/mod.rs
adambezecny: nice! i will redo based on final decision whether **_server_allow_unmasked_** should be **_bool_** or **_Option<bool>_**

agalakhov: My suggestion would be `if self.get_config().server_allow_unmasked.unwrap_or(false)`. But the above one is fine too.

agalakhov:
Thank you!

adambezecny:
I did change Option<bool> to bool, makes more sense and is much lightweight. If you still decide to go with Option I will revert.

src/protocol/mod.rs
```diff
@@ -50,6 +50,11 @@ pub struct WebSocketConfig {
     /// be reasonably big for all normal use-cases but small enough to prevent memory eating
     /// by a malicious user.
     pub max_frame_size: Option<usize>,
+    /// If set to true it will allow the websocket server to accept unmasked frames from client.
+    /// Even though this behaviour is not in compliance with RFC 6455 (which requires the server
+    /// to close the connection when unmasked frame from client is received) it might be handy in some cases
+    /// as there are existing applications sending unmasked client frames.
```
strohel, line 56: Nit: we should probably mention that default is false. (it's kinda obvious, but documentation should probably state even the obvious things)

strohel (commented):
Looks good! I have added one little nit. But take this as an Approve from me - though I'll let @agalakhov / @application-developer-DA actually approve this instead.

src/protocol/mod.rs
adambezecny: documentation improved.

src/protocol/mod.rs
```diff
@@ -50,6 +50,13 @@ pub struct WebSocketConfig {
     /// be reasonably big for all normal use-cases but small enough to prevent memory eating
     /// by a malicious user.
     pub max_frame_size: Option<usize>,
+    /// If set to true it will allow the websocket server to accept unmasked frames from client.
+    /// Even though this behaviour is not in compliance with RFC 6455 (which requires the server
+    /// to close the connection when unmasked frame from client is received) it might be handy in some cases
+    /// as there are existing applications sending unmasked client frames. By default (i.e. unless not explicitly specified)
+    /// this flag is set to false which means violation of RFC 6455 is not allowed and unmasked client frames will be rejected
+    ///by websocket server.
+    pub server_allow_unmasked: bool,
```
application-developer-DA, line 59: A couple minor remarks:

1. I would call it `accept_unmasked_frames` (a bit more descriptive name), what do you think?
2. There is a space missing after the `///` at the last line.
3. I would change the description a bit, text below:

```
/// When set to `true`, the server will accept and handle unmasked frames
/// from the client. According to the RFC 6455, the server must close the
/// connection to the client in such cases, however it seems like there are
/// some popular libraries that are sending unmasked frames, ignoring the RFC.
/// By default this option is set to `false`, i.e. according to RFC 6455.
```

src/protocol/mod.rs
```diff
@@ -449,9 +457,14 @@ impl WebSocketContext {
                     } else {
                         // The server MUST close the connection upon receiving a
                         // frame that is not masked. (RFC 6455)
-                        return Err(Error::Protocol(
-                            "Received an unmasked frame from client".into(),
-                        ));
+                        // The only exception here is if the user explicitly accepts given
+                        // stream (by tungstenite::server::accept_with_config or tungstenite::server::accept_hdr_with_config)
+                        // with WebSocketConfig.server_allow_unmasked set to true
+                        if self.get_config().server_allow_unmasked == false {
```
application-developer-DA, line 463: We do not need to use `get_config()` as `config` is the field of the `self`.

Additionally `self.server_allow_unmasked == false` is the same as `!self.server_allow_unmasked`.

src/protocol/mod.rs
```diff
@@ -449,9 +457,14 @@ impl WebSocketContext {
                     } else {
```
application-developer-DA, line 457: I think the condition below could be use in the `else if` block.

src/protocol/mod.rs
```diff
@@ -449,9 +457,14 @@ impl WebSocketContext {
...
                         // The server MUST close the connection upon receiving a
                         // frame that is not masked. (RFC 6455)
-                        return Err(Error::Protocol(
-                            "Received an unmasked frame from client".into(),
-                        ));
+                        // The only exception here is if the user explicitly accepts given
+                        // stream (by tungstenite::server::accept_with_config or tungstenite::server::accept_hdr_with_config)
+                        // with WebSocketConfig.server_allow_unmasked set to true
```
application-developer-DA, line 462: I would not mention the `accept_with_config()` and `accept_hdr_with_config()` explicitly, I think the option and its usage would be self-explanatory at this place (after all it's documented in the config).

application-developer-DA (changes requested):
I made some remarks, but apart from that it looks fine (did not check the example though).

@strohel , @agalakhov , I would like to mention that I'm actually a bit skeptical regarding https://github.com/snapview/tokio-tungstenite/issues/138 , because it's an attempt to violate the RFC, which potentially allows wrong software to work as if it were written properly, giving the developers of the libraries that violate RFC less inducement to fix their code. In a long run that "promotes" the usage of legacy (which in some cases get dangerous over years, introducing security and other issues, not to mention that it makes things a bit more complex as "working properly" starts to be treated as "buggy", while "violating rules" is "it's fine, we've been always doing so").

That's the only "generally philosophical concern" that I have.

src/protocol/mod.rs
adambezecny: comments incorporated, including renaming the flag to _accept_unmasked_frames_ as suggested. I think _accept_unmasked_frames_ is more explanatory in terms of action but less explanatory in terms of who is doing the action. The original name (server_allow_unmasked) was the opposite case. Ideal would be _server_accept_unmasked_frames_ but that is already too long.

src/protocol/mod.rs
application-developer-DA: Can we still make it part of this `if`? :)
I.e. something like:

```
} else if !self.config.accept_unmasked_frames {
    return Err(....
}
```

src/protocol/mod.rs
application-developer-DA: > but less explanatory in terms of who is doing the action

For the client the option "accept unmasked frames" does not quite make sense, because the client always receives unmasked frames.

application-developer-DA (changes requested):
Some minor change is required, but apart from that I don't have any other remarks.

src/protocol/mod.rs
adambezecny: just to confirm before I push it. Something like below?

```
                    if frame.is_masked() {
                        // A server MUST remove masking for data frames received from a client
                        // as described in Section 5.3. (RFC 6455)
                        frame.apply_mask()
                    } else if !self.config.accept_unmasked_frames {
                        // The server MUST close the connection upon receiving a
                        // frame that is not masked. (RFC 6455)
                        // The only exception here is if the user explicitly accepts given
                        // stream by setting WebSocketConfig.accept_unmasked_frames to true
                        return Err(Error::Protocol(
                            "Received an unmasked frame from client".into(),
                        ));
                    }
```

application-developer-DA: Yes, looks correct.

adambezecny: done. now it is perfect:)

application-developer-DA:
> I made some remarks, but apart from that it looks fine (did not check the example though).
> 
> @strohel , @agalakhov , I would like to mention that I'm actually a bit skeptical regarding [snapview/tokio-tungstenite#138](https://github.com/snapview/tokio-tungstenite/issues/138) , because it's an attempt to violate the RFC, which potentially allows wrong software to work as if it were written properly, giving the developers of the libraries that violate RFC less inducement to fix their code. In a long run that "promotes" the usage of legacy (which in some cases get dangerous over years, introducing security and other issues, not to mention that it makes things a bit more complex as "working properly" starts to be treated as "buggy", while "violating rules" is "it's fine, we've been always doing so").
> 
> That's the only "generally philosophical concern" that I have.

So, apart from that, I don't have further remarks at the moment. As far as I understood, you guys were for this change in general. If so, it's an approve from my side.

application-developer-DA (approved):
Thanks for addressing my comments ;)

strohel (approved):

adambezecny:
hi guys, any ETA with respect to final approval of this PR? Our project is advancing well and it would be good for us to use official version of your library instead of forked one.

PR #166, opened 2021-01-02
Add support for rustls as TLS backend

### LINKED ISSUE #89: Switch from native-tls to rustls

dnaka91:
This PR adds `rustls` as an alternative to the current `native-tls`. In addition, I changed the default feature set to not include any TLS implementation which is a breaking change as people have to add the TLS feature explicitly now.

I'm not sure what way you prefer to have the feature set so I will adjust to the way you prefer once you let me know.

Resolves #89.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,9 +17,9 @@ all-features = true
 
 [features]
 default = []
-use-native-tls = ["native-tls"]
-use-native-tls-vendored = ["use-native-tls", "native-tls/vendored"]
-use-rustls = ["rustls", "webpki", "webpki-roots"]
+native-tls = ["native-tls-crate"]
+native-tls-vendored = ["native-tls", "native-tls-crate/vendored"]
+rustls-tls = ["rustls", "webpki", "webpki-roots"]
 
 [dependencies]
 base64 = "0.13.0"
@@ -35,8 +35,9 @@ thiserror = "1.0.23"
 url = "2.1.0"
 utf-8 = "0.7.5"
 
-[dependencies.native-tls]
+[dependencies.native-tls-crate]
 optional = true
+package = "native-tls"
 version = "0.2.3"
 
 [dependencies.rustls]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,8 +17,9 @@ features = ["native-tls"]
 
 [features]
 default = []
-native-tls-vendored = ["native-tls", "native-tls/vendored"]
-rustls-tls = ["rustls", "webpki", "webpki-roots"]
+use-native-tls = ["native-tls"]
+use-native-tls-vendored = ["use-native-tls", "native-tls/vendored"]
+use-rustls = ["rustls", "webpki", "webpki-roots"]
 
 [dependencies]
 base64 = "0.13.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ edition = "2018"
 all-features = true
 
 [features]
-default = []
+default = ["native-tls"]
 native-tls = ["native-tls-crate"]
 native-tls-vendored = ["native-tls", "native-tls-crate/vendored"]
 rustls-tls = ["rustls", "webpki", "webpki-roots"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ version = "0.13.0"
 edition = "2018"
 
 [package.metadata.docs.rs]
-features = ["native-tls"]
+all-features = true
 
 [features]
 default = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,10 +12,13 @@ repository = "https://github.com/snapview/tungstenite-rs"
 version = "0.11.1"
 edition = "2018"
 
+[package.metadata.docs.rs]
+features = ["native-tls"]
+
 [features]
-default = ["tls"]
-tls = ["native-tls"]
-tls-vendored = ["native-tls", "native-tls/vendored"]
+default = []
+native-tls-vendored = ["native-tls", "native-tls/vendored"]
+rustls-tls = ["rustls", "webpki", "webpki-roots"]
 
 [dependencies]
 base64 = "0.13.0"
@@ -33,6 +36,18 @@ utf-8 = "0.7.5"
 [dependencies.native-tls]
 optional = true
 version = "0.2.3"
+
+[dependencies.rustls]
+optional = true
+version = "0.19.0"
+
+[dependencies.webpki]
+optional = true
+version = "0.21.4"
+
+[dependencies.webpki-roots]
+optional = true
+version = "0.21.0"
 
 [dev-dependencies]
 env_logger = "0.8.1"
--- a/src/client.rs
+++ b/src/client.rs
@@ -26,17 +26,20 @@ mod encryption {
     /// TCP stream switcher (plain/TLS).
     pub type AutoStream = StreamSwitcher<TcpStream, TlsStream<TcpStream>>;
 
-    use crate::{error::Result, stream::Mode};
+    use crate::{
+        error::{Result, TlsError},
+        stream::Mode,
+    };
 
     pub fn wrap_stream(stream: TcpStream, domain: &str, mode: Mode) -> Result<AutoStream> {
         match mode {
             Mode::Plain => Ok(StreamSwitcher::Plain(stream)),
             Mode::Tls => {
-                let connector = TlsConnector::builder().build()?;
+                let connector = TlsConnector::builder().build().map_err(TlsError::Native)?;
                 connector
                     .connect(domain, stream)
                     .map_err(|e| match e {
-                        TlsHandshakeError::Failure(f) => f.into(),
+                        TlsHandshakeError::Failure(f) => TlsError::Native(f).into(),
                         TlsHandshakeError::WouldBlock(_) => {
                             panic!("Bug: TLS handshake not blocked")
                         }
@@ -58,7 +61,10 @@ mod encryption {
     /// TCP stream switcher (plain/TLS).
     pub type AutoStream = StreamSwitcher<TcpStream, StreamOwned<ClientSession, TcpStream>>;
 
-    use crate::{error::Result, stream::Mode};
+    use crate::{
+        error::{Result, TlsError},
+        stream::Mode,
+    };
 
     pub fn wrap_stream(stream: TcpStream, domain: &str, mode: Mode) -> Result<AutoStream> {
         match mode {
@@ -70,7 +76,7 @@ mod encryption {
 
                     Arc::new(config)
                 };
-                let domain = DNSNameRef::try_from_ascii_str(domain)?;
+                let domain = DNSNameRef::try_from_ascii_str(domain).map_err(TlsError::Dns)?;
                 let client = ClientSession::new(&config, domain);
                 let stream = StreamOwned::new(client, stream);
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -129,8 +129,8 @@ use crate::{
 /// custom stream, call `client` instead.
 ///
 /// This function uses `native_tls` or `rustls` to do TLS depending on the feature flags enabled. If
-/// you want to use other TLS libraries, use `client` instead. There is no need to enable the "tls"
-/// feature if you don't call `connect` since it's the only function that uses native_tls or rustls.
+/// you want to use other TLS libraries, use `client` instead. There is no need to enable any of
+/// the `*-tls` features if you don't call `connect` since it's the only function that uses them.
 pub fn connect_with_config<Req: IntoClientRequest>(
     request: Req,
     config: Option<WebSocketConfig>,
@@ -197,8 +197,8 @@ pub fn connect_with_config<Req: IntoClientRequest>(
 /// custom stream, call `client` instead.
 ///
 /// This function uses `native_tls` or `rustls` to do TLS depending on the feature flags enabled. If
-/// you want to use other TLS libraries, use `client` instead. There is no need to enable the "tls"
-/// feature if you don't call `connect` since it's the only function that uses native_tls or rustls.
+/// you want to use other TLS libraries, use `client` instead. There is no need to enable any of
+/// the `*-tls` features if you don't call `connect` since it's the only function that uses them.
 pub fn connect<Req: IntoClientRequest>(request: Req) -> Result<(WebSocket<AutoStream>, Response)> {
     connect_with_config(request, None, 3)
 }
--- a/src/client.rs
+++ b/src/client.rs
@@ -16,10 +16,10 @@ use crate::{
     protocol::WebSocketConfig,
 };
 
-#[cfg(feature = "use-native-tls")]
+#[cfg(feature = "native-tls")]
 mod encryption {
-    pub use native_tls::TlsStream;
-    use native_tls::{HandshakeError as TlsHandshakeError, TlsConnector};
+    pub use native_tls_crate::TlsStream;
+    use native_tls_crate::{HandshakeError as TlsHandshakeError, TlsConnector};
     use std::net::TcpStream;
 
     pub use crate::stream::Stream as StreamSwitcher;
@@ -47,7 +47,7 @@ mod encryption {
     }
 }
 
-#[cfg(all(feature = "use-rustls", not(feature = "use-native-tls")))]
+#[cfg(all(feature = "rustls-tls", not(feature = "native-tls")))]
 mod encryption {
     use rustls::ClientConfig;
     pub use rustls::{ClientSession, StreamOwned};
@@ -80,7 +80,7 @@ mod encryption {
     }
 }
 
-#[cfg(not(any(feature = "use-native-tls", feature = "use-rustls")))]
+#[cfg(not(any(feature = "native-tls", feature = "rustls-tls")))]
 mod encryption {
     use std::net::TcpStream;
 
@@ -116,7 +116,7 @@ use crate::{
 /// equal to calling `connect()` function.
 ///
 /// The URL may be either ws:// or wss://.
-/// To support wss:// URLs, feature `use-native-tls` or `use-rustls` must be turned on.
+/// To support wss:// URLs, feature `native-tls` or `rustls-tls` must be turned on.
 ///
 /// This function "just works" for those who wants a simple blocking solution
 /// similar to `std::net::TcpStream`. If you want a non-blocking or other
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -12,10 +12,13 @@ repository = "https://github.com/snapview/tungstenite-rs"
 version = "0.11.1"
 edition = "2018"
 
+[package.metadata.docs.rs]
+features = ["native-tls"]
```
application-developer-DA, line 16: I'm thinking what `features` do we want to have here in the docs. Perhaps it would make sense to have `rustls-tls` instead? But then maybe it would make more sense to use `rustls-tls` as a default feature? @agalakhov, @strohel what do you think?

application-developer-DA (approved):
Thanks, in general looks good! We need to decide on the features (the point that you also mentioned) and then we can merge I think.

dnaka91:
@application-developer-DA Thanks for the quick review!

There is another thing I stumbled upon while preparing the rustls changes for `tokio-tungstenite` and would like your input. Cargo doesn't allow to have features and dependencies with the same name. Here we only need `native-tls` as dependency for the same named feature and it works. In `tokio-tungstenite` though, there are multiple dependencies involved to enable the native TLS feature and therefore it can't be named `native-tls` and I went with the inverse `tls-native-tls` and `tls-rustls`.
As said, currently it just works here because we only need that single dependency but for future purposes it may be good to avoid having the same name for dependency and feature as we can't add more dependencies to that feature in the future without renaming (there is an exception though).

I saw that `reqwest` is doing another approach by renaming the dependency instead. So the feature is `native-tls` and the same named depedency was renamed to `native-tls-crate` to avoid the name clash. That looks like this for the dependency:
```toml
native-tls-crate = { version = "0.2", optional = true, package = "native-tls" }
```

This will rename it in the code as well so there will be `use native_tls_crate::...`.

With all that said, what would you prefer?
1. Just keep it as is and maybe use the same technique as it `reqwest` in the future, whenever the need for that comes.
2. Rename to maybe `tls-native-tls` (or just `tls-native`) and `tls-rustls`.

And in regards to `tokio-tungstenite` as we have multiple crates to enable native TLS, what would you like me to use there?

application-developer-DA:
Hm, good point @dnaka91. The approach which `reqwest` uses looks tempting at the first sight, but I think using a "custom" name for the crate in the code is not the nicest thing probably, so I personally would prefer different name for features rather than renaming of the crates that we're fetching as dependencies. What do you think of feature names like `use-native-tls` or `use-rustls`? (IMHO they would be self-descriptive and easy to understand)

dnaka91:
I think the same. Would probably be confusing as well if common crate names suddenly have another name within the project only (especially for people who don't know about that renaming feature).

Using a `use-*` prefix sounds good and I think I saw 1 or 2 other crates with a similar naming. Some even did something like `*-backend` or even longer named approaches.

Only one concern I have with the `use-*` naming is when crates don't have `tls` in their name. For example `openssl` could be used for TLS as well in the future (I think same effect as `native-tls` on Linux but another story on Windows and MacOS). In that case it would be a little unclear what `use-openssl` is actually used for. The TLS connection, some key generation or other cryptographic functions that may be something not related to TLS?
That would only be the case for when there is any plan to add `openssl` for TLS in the future.... I personally thing native-tls and rustls are sufficient but other people may thing differently.

So besides that, I really like the `use-*` naming actually.

dnaka91:
Well, maybe I overthink things a little bit here. So I just updated to use the `use-*` prefix.
So far nobody asked for a explicit `openssl` variant for TLS and probably okay to wait to add it until somebody actually needs it (also can use name it `use-openssl-tls` in that case I guess).

src/error.rs
```diff
@@ -40,9 +47,15 @@ pub enum Error {
     /// Input-output error. Apart from WouldBlock, these are generally errors with the
     /// underlying connection and you should probably consider them fatal.
     Io(io::Error),
-    #[cfg(feature = "tls")]
+    #[cfg(feature = "use-native-tls")]
     /// TLS error
     Tls(tls::Error),
+    #[cfg(feature = "use-rustls")]
```
sdroege, line 53: This is suboptimal. cargo features are supposed to be additive but enabling both here would cause compiler errors.

You can work around that by making the conditions a bit more complicated so that only one is actually used if both are enabled, but that would still cause problems if in the whole dependency chain one crate wants one feature and the other wants the other one.

The only proper solution to this is to reorganize things in a way that both features can be successfully enabled at once.

dnaka91: That could indeed be a problem, didn't think about that. Most of the time you want either one of them but in bigger projects where you have multiple client to some 3rd party APIs that all use tungstenite, not all of them might give the option to select the TLS library.
How likely is that to happen though? Most libraries allow to pick or just transfer the decision to the user.

I saw that you use a fallback mechanism in `async-tungstenite` where you default to `native-tls` or `rustls` depending on what features are enabled.

Best would be to be explicit I guess but that would put an extra burden on the user for situations that may not happen too often. What I mean is we couldn't simple accept `wss://` URLs in `connect` anymore but would always require to pass extra information about the TLS implementation to use. That could be either a new parameter or have a separate `connect_tls` function maybe.
The disadvantage is that you would always have to pass the parameter like `connect_tls("wss://example.com", TlsConnector::Rustls);` even though you mostly only enable one TLS backend so you wouldn't really have to pass it and it only solves the rare case of having both backends enabled.

Maybe there could be a middleground where `connect` simply falls back to a pre-defined default, like always picking `native-tls` if both are enabled and add a new field to the config for `connect_with_config` that allows to be explicit.
That could be the most elegant version eventually, but can cause hidden side-effects if you rely on specific features of a TLS connector (for example `rustls` doesn't support doesn't support IP SANs yet) as the default may be different if both features happen to be enabled in user's crate and you can only circumvent it by always be explicit (and you may not even know about that detail).

That are my thoughts on this so far. What do you think @application-developer-DA ?

sdroege: >  How likely is that to happen though? Most libraries allow to pick or just transfer the decision to the user.

In a bigger application you might have two dependencies that as part of their job talk to some websocket API, and both could potentially select different TLS libraries for arbitrary reasons. It's not very likely but it's also very annoying if it happens :)

The main problem here is API surface that is changing in incompatible ways depending on the features selected.

dnaka91: So I spend quite a few hours now shelling out a new variant of this changes, allowing both features to coexist and have a nice switch between no TLS at all and one or more TLS features enabled.

There are still several rough edges but it's very late for me and I will stop at this point for today. Would be great if you can take a look and tell me what you think @sdroege : https://github.com/snapview/tungstenite-rs/compare/master...dnaka91:rustls-multi-feature?expand=1

As you will see the changes are rather large to make it work even though I tried to keep changes at a minimum. The amount of extra feature flag checks is quite unfortunate as well.

Whenever you have time @application-developer-DA please let me know your opinions on this topic as well.

dnaka91: @sdroege I updated this PR to allow both features at the same time now.

Now if both are enabled, then both errors will be enabled on the `Error` enum and some trait implementations are enabled as well. But for the actual TLS only `native-tls` will be used as auto-fallback. I would personally prefer `rustls` but the fact that `webpki` doesn't support IP SNIs yet makes it more likely to break clients if it would switch over.

The only pitfall currently is that someone might check for TLS errors from `rustls` but will never catch them as the crate would switch over to `native-tls`.

Maybe this could be merged as a first version of `rustls` support and eventually have the explicit variant in the future. What do you think?

Cargo.toml
application-developer-DA: @dnaka91 , was it meant to be `use-native-tls` instead of `native-tls`? :)

src/error.rs
application-developer-DA: @dnaka91 the changes look good. Apart from the issues that @sdroege noticed I don't see any other issues with the PR. Do you plan to implement the suggestions that you mentioned to avoid issues in cases that @sdroege mentioned? I'm asking, so that I know if we can review it considering it to be some sort of a "final state" of the PR so to say :)

dnaka91: Hey @application-developer-DA. So currently this PR should be complete, except I noticed some function docs mention the usage of native-tls that may need some update.

I adjusted the features so that both can be enabled at the same time and if they are then `native-tls` is used, but other things like the `TlsError` enum will have things from both crates so that no sudden compile errors happen.

My other idea was to make TLS usage fully explicit but that would require some bigger API changes. I thought to have `connect` for just plain connections without TLS and without any detection bases on the URL schema. Instead only accept `ws://` and no `wss://` then the `wss://` would only be allowed in a `connect_tls` function that accepts a `Tls` enum where we can pick which one to use. Alternatively have just two functions `connect_tls_native` and `connect_tls_rustls`.

I tried these changes on a separate branch and the differences are quite big (tried something here but this is not complete yet: https://github.com/snapview/tungstenite-rs/compare/master...dnaka91:rustls-multi-feature?expand=1).

src/error.rs
```diff
@@ -6,17 +6,12 @@ use crate::protocol::{frame::coding::Data, Message};
...
-    pub use native_tls::Error;
-}
-
 /// Result type of all Tungstenite library calls.
 pub type Result<T> = result::Result<T, Error>;
 
 /// Possible WebSocket errors.
 #[derive(Error, Debug)]
+#[non_exhaustive]
 pub enum Error {
```
strohel, line 15: While the `#[non_exhaustive]` technically resolves the API incompatibility problems with additivity of features that @sdroege mentioned, I'm not sure whether it is worth it.

It forces downstream crates to add `_ => handle unknown error`, which would be triggered when the crate only handles the "other" Tls error type - which would be most likely unwanted.

It seems that a "proper" solution would to provide a common abstraction of Tls errors. Inside, it could be #[non_exhaustive] enum of possible feature-gated types.

What do you think? Worth it? Note that we already had the API changing based on the `tls` feature flag.

strohel (changes requested):

Cargo.toml
```diff
@@ -12,10 +12,14 @@ repository = "https://github.com/snapview/tungstenite-rs"
...
+
 [features]
-default = ["tls"]
-tls = ["native-tls"]
-tls-vendored = ["native-tls", "native-tls/vendored"]
+default = []
+use-native-tls = ["native-tls"]
+use-native-tls-vendored = ["use-native-tls", "native-tls/vendored"]
+use-rustls = ["rustls", "webpki", "webpki-roots"]
```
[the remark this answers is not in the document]
application-developer-DA: @strohel, but are not `use-[name_of_the_crate]` not self-explanatory names of features? I mean, I don't think that `use-native-tls` or `use-rustls` would be confusing as it's basically like saying "use mylibrary1", "use mylibrary2", i.e. quite a regular English.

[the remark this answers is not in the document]
strohel: Checking other prominent crates in the Rust ecosystem (e.g. tokio), the feature names seem to be formulated as simple nouns, not as verbs/sentences/orders. My main concern here is consistency (across ecosystem; in terms of naming style and also in terms of using the very same names for identical intents), i.e. reduced fragmentation, thus less cognitive load for our downstreams.

I agree that `use-something` sounds naturally in English. But extending the idea, then nearly every feature flag could be named `use-something`, i.e. the `use` would bring little information value (as in namig every struct `SomeStruct`, every enum `SomeEnum` bringing also little information value).

Another rather minor rationale is that `use-rustls` says "Use rustls for something", while `rustls-tls` says "(enable) TLS through rustls", arguably slightly more information. (this was already mentioned earlier in the debate)

strohel (changes requested):
Once we reach consensus on the feature flag name, please also add an entry to CHANGELOG.md under version 0.13.

Speaking of changelog, it should probably express that 0.13 has not yet been released as advised in https://keepachangelog.com/ - but that's out of scope of this PR and feel free to ignore that.

Cargo.toml
strohel: > But then maybe it would make more sense to use rustls-tls as a default feature?

I think that would be a good future goal, but @dnaka91 mentioned
> the fact that webpki doesn't support IP SNIs yet makes it more likely to break clients if it would switch over.

So I would say not yet.

Cargo.toml
```diff
@@ -12,10 +12,14 @@ repository = "https://github.com/snapview/tungstenite-rs"
...
 edition = "2018"
 
+[package.metadata.docs.rs]
+features = ["native-tls"]
+
 [features]
-default = ["tls"]
-tls = ["native-tls"]
-tls-vendored = ["native-tls", "native-tls/vendored"]
+default = []
```
strohel, line 19: I would suggest to enable TLS support by default. (and as discussed above, the default should lead to native TLS for now) I'm a big fan of sane defaults, and TLS enabled is arguable a very sane default in 2021.

People with special needs can still use `default-features = false`.

strohel (commented):
(sorry for submitting the review in 3 parts)

Cargo.toml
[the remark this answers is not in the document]
dnaka91: I personally prefer the Reqwest way with `rustls-tls` and `native-tls` as well. For one reason just for the consistency and as you already explained so nicely because `use-rustls` doesn't say for what it's used whereas `rustls-tls` clearly says it's used for TLS.

What is the final decision for this?

src/error.rs
dnaka91: You mean to make the error opaque by hiding the real errors behind it and just have a single TLS error?
Or do you mean more of detailing out the error into as many details as each TLS error from the crates provide so we can convert to our own TLS error implementation but still give the user the opportunity to match on the exact TLS problem?

Cargo.toml
dnaka91: I myself prefer the _tokio-style_ having nothing enabled by default and make it more of a _pick what you need_.

But I don't mind enabling `native-tls` as default, so I'll update this once the feature flag naming is resolved.

Cargo.toml
dnaka91: Oh thanks, I forgot to update this. It was initially set like this because in the first version there could only be one of the two feature flags enabled at the same times.

Let's switch to enable all features, now that they can co-exist.

dnaka91:
@strohel regarding the CHANGELOG.md I believe it's not fully conform to the https://keepachangelog.com spec and I would like to update that in a separate PR. For now I'll just add to the 0.13 as suggested 🙇

src/error.rs
strohel: I believe this thread can be resolved now, as the features are now additive (with other comments dealing with some corner-cases). I don't believe the bigger changes for explicit TLS usage are needed now.

src/error.rs
strohel: I see I was vague in my original comment, so I'd better express that with code:
```rust
/// Possible WebSocket errors. [NOTE: exhaustive]
#[derive(Error, Debug)]
pub enum Error {
    (...)
    Io(#[from] io::Error),
    /// TLS error. [NOTE: enabled unconditionally even if no tls feature is enabled for feature-agnostic API surface]
    #[error("TLS error: {0}")]
    Tls(#[from] TlsError),
    /// - When reading: buffer capacity exhausted.
    (...)
}

/// TLS errors.
///
/// Note that even if you enable only the rustls-based TLS support, the error at runtime could still
/// be `Native`, as another crate in the dependency graph may enable native TLS support.
#[derive(Error, Debug)]
#[non_exhaustive]
pub enum TlsError {
    /// Native TLS error.
    #[cfg(feature = "use-native-tls")]
    #[error("native-tls error: {0}")]
    Native(#[from] native_tls::Error),
    /// Rustls error.
    #[cfg(feature = "use-rustls")]
    #[error("rustls error: {0}")]
    Rustls(#[from] rustls::TLSError),
    /// DNS name resolution error.
    #[cfg(feature = "use-rustls")]
    #[error("Invalid DNS name: {0}")]
    Dns(#[from] webpki::InvalidDNSNameError),
}
```

I've checked what abstraction we provide on top of https://docs.rs/rustls/0.19.0/rustls/enum.TLSError.html and https://docs.rs/native-tls/0.2.7/native_tls/struct.Error.html, but native_tls::Error doesn't provide anything more than `Debug + Display + Error`.

Note that this is a brainstorm rather than a serious suggestion. It depends on what you (and possibly @application-developer-DA, @jxs, @sdroege) think -- whether this is worth it or not.

dnaka91: Thanks for all the effort preparing a code sample.

I totally understand now what you mean. That's a good idea to keep the `non_exhaustive` usage limited.

I'll prepare the changes together with your other suggestions tomorrow.

Cargo.toml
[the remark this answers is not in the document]
strohel: I had a chat with @application-developer-DA, he'll survive `rustls-tls` and `native-tls` ( :-) ), please go ahead.

dnaka91:
@strohel Okay I hope I didn't forget any point from our review.
 - All features enabled for docs.rs.
 - Feature flags renamed to `*-tls`.
 - `native-tls` included in the default feature set.
 - TLS related errors moved to a sub-error.
 - Changelog entry added.

I tested all changes with `cargo-hack` to make sure that all feature flag combinations compile and tests pass.

src/client.rs
```diff
@@ -83,15 +122,15 @@ use crate::{
...
 /// This function "just works" for those who wants a simple blocking solution
 /// similar to `std::net::TcpStream`. If you want a non-blocking or other
 /// custom stream, call `client` instead.
 ///
-/// This function uses `native_tls` to do TLS. If you want to use other TLS libraries,
-/// use `client` instead. There is no need to enable the "tls" feature if you don't call
-/// `connect` since it's the only function that uses native_tls.
+/// This function uses `native_tls` or `rustls` to do TLS depending on the feature flags enabled. If
+/// you want to use other TLS libraries, use `client` instead. There is no need to enable the "tls"
+/// feature if you don't call `connect` since it's the only function that uses native_tls or rustls.
```
strohel, line 133: Nitty clarification (feel free to adapt):
```suggestion
/// you want to use other TLS libraries, use `client` instead. There is no need to enable any of
/// the `*-tls` features if you don't call `connect` since it's the only function that uses them.
```
Once more occurrence below, as we have some docs duplication.

strohel (approved):
This is perfect, @dnaka91! Thanks for very diligent update, TIL about `cargo-hack`, useful.

I have one little suggestion about docs, and we've merged a change in the mean time that caused a merge conflict in CHANGELOG.md - please rebase.

After that, this is ripe for merging. To get another pair of eyeballs on the final version, I'll lett @agalakhov / @application-developer-DA / @jxs merge or approve this.

strohel:
> After that, this is ripe for merging. To get another pair of eyeballs on the final version, I'll lett @agalakhov / @application-developer-DA / @jxs merge or approve this.

Forgot to say, to anyone merging this incl. me: please squash all the commits into a single commit when when merging.

src/client.rs
dnaka91: Thanks for the suggestion. That is much clearer 🙇‍♂️

dnaka91:
Okay, I merged in the latest changes from master (should make no difference from rebasing if the PR is squash merged) and adjusted the docs as you mentioned.

After this is merged let's head over to `tokio-tungstenite` https://github.com/snapview/tokio-tungstenite/pull/147 😉. Well, first there needs to be a new release of this crate before that one can be merged.

application-developer-DA (approved):

PR #168, opened 2021-01-04
Create specific error types for protocol, URL, and capacity errors

WiredSound:
The current tungstenite error system doesn't allow for easily matching specific protocol, URL, and capacity errors due to use of string messages to distinguish different error subtypes. As such, I've created distinct enums for the different protocol, URL, and capacity errors that may occur.

To give an example of how this would be useful, imagine a scenario where one wants to check if an operation failed due to the remote host closing the connection without performing the closing handshake. Currently, one may write `Err(Error::Protocol(vioation)) if vioation.contains("closing handshake") => { ... }` or similar. With my new change, they now may write `Err(Error::Protocol(ProtocolErrorType::ResetWithoutClosingHandshake)) => { ... }` instead. This is both more efficient as it avoids comparing strings while also allowing the error message to be changed without breaking existing pattern matching.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # 0.13.0
 
-- Add `thiserror` to dependencies.
-- Modify `Error` to be implemented using the `thiserror::Error` derive macro.
 - Add `CapacityError`, `UrlError`, and `ProtocolError` types to represent the different types of capacity, URL, and protocol errors respectively.
 - Modify variants `Error::Capacity`, `Error::Url`, and `Error::Protocol` to hold the above errors types instead of string error messages.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 0.13.0
+
+- Add `thiserror` to dependencies.
+- Modify `Error` to be implemented using the `thiserror::Error` derive macro.
+- Add `CapacityError`, `UrlError`, and `ProtocolError` types to represent the different types of capacity, URL, and protocol errors respectively.
+- Modify variants `Error::Capacity`, `Error::Url`, and `Error::Protocol` to hold the above errors types instead of string error messages.
+
 # 0.12.0
 
 - Add facilities to allow clients to follow HTTP 3XX redirects.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ readme = "README.md"
 homepage = "https://github.com/snapview/tungstenite-rs"
 documentation = "https://docs.rs/tungstenite/0.12.0"
 repository = "https://github.com/snapview/tungstenite-rs"
-version = "0.12.0"
+version = "0.13.0"
 edition = "2018"
 
 [features]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ rand = "0.8.0"
 sha-1 = "0.9"
 url = "2.1.0"
 utf-8 = "0.7.5"
+thiserror = "1.0.23"
 
 [dependencies.native-tls]
 optional = true
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,9 +1,10 @@
 //! Error handling.
 
-use std::{error::Error as ErrorTrait, fmt, io, result, str, string};
+use std::{fmt, io, result, str, string};
 
 use crate::protocol::{frame::coding::Data, Message};
 use http::Response;
+use thiserror::Error;
 
 #[cfg(feature = "tls")]
 pub mod tls {
@@ -15,7 +16,7 @@ pub mod tls {
 pub type Result<T> = result::Result<T, Error>;
 
 /// Possible WebSocket errors.
-#[derive(Debug)]
+#[derive(Error, Debug)]
 pub enum Error {
     /// WebSocket connection closed normally. This informs you of the close.
     /// It's not an error as such and nothing wrong happened.
@@ -28,6 +29,7 @@ pub enum Error {
     ///
     /// Receiving this error means that the WebSocket object is not usable anymore and the
     /// only meaningful action with it is dropping it.
+    #[error("Connection closed normally")]
     ConnectionClosed,
     /// Trying to work with already closed connection.
     ///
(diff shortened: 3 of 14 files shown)


application-developer-DA (approved):
Thanks! I had skimmed through the code and it looks like a reasonable change. I remember we even discussed similar ideas with @jxs once we spoke about usage of `thiserror` in `tungstenite`, though it would be a scope of another PR.

application-developer-DA (approved):

Cargo.toml
```diff
@@ -9,7 +9,7 @@ readme = "README.md"
 homepage = "https://github.com/snapview/tungstenite-rs"
 documentation = "https://docs.rs/tungstenite/0.11.1"
 repository = "https://github.com/snapview/tungstenite-rs"
-version = "0.11.1"
+version = "0.12.0"
```
strohel, line 12: The master has been already bumped to 0.12.0, please rebase and make this a bump from 0.12.0 to 0.13.0. (I believe a bump is necessary as this commit introduces an incompatible change to the library API).

Cargo.toml
```diff
@@ -27,7 +27,7 @@ input_buffer = { version = "0.4.0", git = "https://github.com/snapview/input_buf
 log = "0.4.8"
 rand = "0.8.0"
 sha-1 = "0.9"
-url = "2.1.0"
+url = "2.2.0"
```
strohel, line 30: Does the changes in this PR _require_ (to build properly, or to work proberly) to use at lest url 2.2.0?

If not, I'd suggest not bumping this, cargo should happily pull url-2.2.0 even if we specify 2.1.0 (that's different to 0.x.y crates, where even minor version is considered incompatible).

src/error.rs
```diff
@@ -146,8 +146,190 @@ impl From<tls::Error> for Error {
...
+        match self {
+            CapacityErrorType::TooManyHeaders => write!(f, "Too many headers"),
+            CapacityErrorType::HeaderTooLong => write!(f, "Header too long"),
+            CapacityErrorType::MessageTooLong { size, max_size } => {
+                write!(f, "Message too long: {} > {}", size, max_size)
+            }
+            CapacityErrorType::TcpBufferFull => write!(f, "Incoming TCP buffer is full"),
+        }
+    }
+}
```
strohel, line 184: I would slightly prefer using `thiserror` to derive this. It's distinctive advantage is that it doesn't end up in the API footprint of the library, it is just "an internal implementation helper".

Because of that, I think it would be in-scope of this PR, as it would allow us not to write (and then change) this code at all.

src/error.rs
```diff
@@ -41,20 +41,20 @@ pub enum Error {
...
     /// Protocol violation.
-    Protocol(Cow<'static, str>),
+    Protocol(ProtocolErrorType),
     /// Message send queue full.
     SendQueueFull(Message),
     /// UTF coding error
     Utf8,
     /// Invalid URL.
-    Url(Cow<'static, str>),
+    Url(UrlErrorType),
```
strohel, line 57: Idea: should we call the payload types just `CapacityError`, `ProtocolError`, `UrlError`? The word `Type` inside the names doesn't seem to add a lot of information (the fact that these are enums already says that it is "a type of something").

strohel (changes requested):
Hi @WiredSound, thanks for the PR, I like what it achieves!

There are some changes needed given that we've already released 0.12. Also, @application-developer-DA has in the mean time introduced a `CHANGELOG.md` file - could you please add an entry for this change to the new file? Please rebase to do so.

I also have some suggestions on the code.

Cargo.toml
WiredSound: Apologies, this was a result of my misunderstanding of how Cargo decides which dependency version to fetch - I've changed it back now!

src/error.rs
WiredSound: Thank you for making me of `thiserror` - very handy!

src/error.rs
WiredSound: The reason I didn't originally call them just `...Error` was because I hadn't actually implemented the `std::error::Error` trait for each of them. Now that I'm using `thiserror` which automatically implements the error trait, this seems like a good idea.

WiredSound:
Thank you for your suggestions, @strohel! Hopefully all should be good to merge now?

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 0.13.0
+
+- Add `thiserror` to dependencies.
+- Modify `Error` to be implemented using the `thiserror::Error` derive macro.
```
strohel, line 4: Thanks for adding ChangeLog entries. I'd remove these first two as they are rather implementation details.

src/protocol/frame/mod.rs
```diff
@@ -266,9 +269,9 @@ mod tests {
...
         let raw = Cursor::new(vec![0x82, 0x07, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07]);
         let mut sock = FrameSocket::new(raw);
-        assert_eq!(
-            sock.read_frame(Some(5)).unwrap_err().to_string(),
-            "Space limit exceeded: Message length too big: 7 > 5"
-        );
+        match sock.read_frame(Some(5)) {
+            Err(Error::Capacity(CapacityError::MessageTooLong { size: 7, max_size: 5 })) => {}
+            _ => panic!(),
+        }
```
strohel, line 275: nit: these assertions could be AFAIK expressed more succinctly by
```suggestion
        assert!(matches!(sock.read_frame(Some(5)), Err(Error::Capacity(CapacityError::MessageTooLong { size: 7, max_size: 5 }))));
```
I believe that `cargo clippy --all-targets` would suggest this, too?

strohel (changes requested):
Nice, thanks. I've left some last nitpicks, otherwise this looks good to go - good job!

src/protocol/frame/mod.rs
WiredSound: That's weird, clippy didn't pick up on that even with `--all-targets`. I wasn't aware of `matches!` - thanks for pointing that out.

src/protocol/mod.rs
```diff
@@ -715,20 +712,22 @@ mod tests {
...
         let mut socket = WebSocket::from_raw_socket(WriteMoc(incoming), Role::Client, Some(limit));
-        assert_eq!(
-            socket.read_message().unwrap_err().to_string(),
-            "Space limit exceeded: Message too big: 7 + 6 > 10"
-        );
+
+        match socket.read_message() {
+            Err(Error::Capacity(CapacityError::MessageTooLong { size: 13, max_size: 10 })) => {}
+            _ => panic!(),
+        }
```
strohel, line 719: Oh, I forgot to mention it in my previous review - there were in total 3 instances where matches!() should be used - this one plus one below.

WiredSound: Ah, my apologies, all fixed now.

strohel (approved):
Looks good, thanks! Given that @application-developer-DA has already approved, I'll merge this.

njam:
:+1:  nice change

PR #174, opened 2021-02-06
Expose `handshake::derive_accept_key` as public API.

de-vri-es:
This PR renames `convert_key` to `derive_accept_key` and exposes it as a public API. This makes it easier to manually perform a handshake before passing the TCP stream to `WebSocket::from_raw_socket`.

Also, thanks for this crate. It's very useful :)


agalakhov:
Applied, thanks.

PR #192, opened 2021-04-04
Remove thiserror dependency

c410-f3r:
Removes two transient dependencies for downstream libraries.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,6 @@ input_buffer = "0.4.0"
 log = "0.4.8"
 rand = "0.8.0"
 sha-1 = "0.9"
-thiserror = "1.0.23"
 url = "2.1.0"
 utf-8 = "0.7.5"
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,16 +1,22 @@
 //! Error handling.
 
-use std::{io, result, str, string};
+mod capacity_error;
+mod protocol_error;
+mod tls_error;
+mod url_error;
 
-use crate::protocol::{frame::coding::Data, Message};
+use crate::protocol::Message;
+pub use capacity_error::CapacityError;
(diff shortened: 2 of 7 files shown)


strohel:
Hi @c410-f3r, thanks for a PR. We've deliberately decided to use thiserror to reduce boilerplate code. But such a decision can be revisited, especially is build time savings would be significant.

Can you please measure build times on your computer, with and without this patch applied? 2 cases, ideally: clean release build `cargo clean; time cargo build --release`; incremental development build: `time cargo build` (drop the first run). Please do multiple runs of each measurement so that we get a sense what the natural variance is. Thanks!

c410-f3r:
Before this PR (1 thread):
```
real    0m27,821s
user    0m24,481s
sys     0m2,461
```

After this PR (1 thread):
```
real    0m20,859s
user    0m18,898s
sys     0m2,019s
```

`cargo build` dependencies went down from `76` to `66`. Don't know why.
It is OK if you guys prefer `thiserror`, my motivation was the addition of more dependencies into downstream libraries. Please, just don't use `anyhow` in the future.

agalakhov:
This looks like reverting the introduction of `thiserror`. Looks like it is the classical "readability vs performance".

application-developer-DA:
> Please, just don't use anyhow in the future.

We don't plan to use `anyhow`. In fact `anyhow` is not meant for libraries as its main purpose is to be used in binary crates. `thiserror` is pretty much the version of `anyhow` for libraries if I'm allowed to generalise it like this.

The difference in build time seems to be negligible (just 10% on my machine, which is around 2,5 second in my particular setup), so I would prefer the version with `thiserror` as it allows to have a cleaner code which is easy to read and maintain.

c410-f3r:
Closed due to the above review

PR #198, opened 2021-04-16
Replace the utf-8 library with std::str::from_utf8 usage

### LINKED ISSUE #197: Consider replacement for the utf-8 crate
The `utf-8` crate is somewhat unmaintained (https://github.com/SimonSapin/rust-utf8/pull/6) and it lacks license files though the Apache-2.0 license requires it (https://github.com/SimonSapin/rust-utf8/issues/3). I would like to use the `tungstenite-rs` crate, but the company I'm working at doesn't permit using of crates without license.

I suppose even `std::str::from_utf8` can be used instead: `std::str::Utf8Error` provides the `valid_up_to` method which can be used to build a utf8 string from parts. This will require a little bit more code though.

stanislav-tkach:
Closes #197.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,6 @@ rand = "0.8.0"
 sha-1 = "0.9"
 thiserror = "1.0.23"
 url = "2.1.0"
-utf-8 = "0.7.5"
 
 [dependencies.native-tls-crate]
 optional = true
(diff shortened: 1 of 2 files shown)


application-developer-DA:
Decided to close as per https://github.com/snapview/tungstenite-rs/issues/197#issuecomment-821313119

PR #200, opened 2021-04-22
Use Autobahn Suite from Docker, symmetrical to snapview/tokio-tungstenite#165

application-developer-DA:
See https://github.com/snapview/tokio-tungstenite/pull/165 for more
details.


strohel (approved):
Cool.

Issue #203, opened 2021-05-05
Remove hard webpki-roots dependency by allowing the user to supply a Rustls `ClientConfig` when using Rustls


briansmith:
Currently in the code we have:
```
            Mode::Tls => {
                let config = {
                    let mut config = ClientConfig::new();
                    config.root_store.add_server_trust_anchors(&webpki_roots::TLS_SERVER_ROOTS);

                    Arc::new(config)
                };
```

We need to use `rustls-native-certs` instead of webpki-roots. Instead of adding a feature that just changes the hard-coded configuration to use `rustls-native-certs`, I propose that we extend the Tungstenite's API, when the Rustls feature is enabled, to allow the application to pass in their own `Arc<rustls::ClientConfig>` that Tungstenite will use instead of creating its own. This way, the application can customize all aspects of Rustls configuration, including cipher suites and everything else.

PR #205, opened 2021-05-18
downgrade webpki, use rustls-native-certs instead of webpki-roots

yusdacra:
rustls 0.19 still uses webpki 0.21, and webpki 0.22 contains breaking changes so webpki must be downgraded. This PR also changes it so that rustls-native-certs is used instead of webpki-roots (related #203).

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -72,7 +72,10 @@ mod encryption {
             Mode::Tls => {
                 let config = {
                     let mut config = ClientConfig::new();
-                    config.root_store.add_server_trust_anchors(&webpki_roots::TLS_SERVER_ROOTS);
+                    config.root_store = match rustls_native_certs::load_native_certs() {
+                        Ok(store) => store,
+                        Err((_, err)) => Err(err)?,
+                    };
 
                     Arc::new(config)
                 };
--- a/src/client.rs
+++ b/src/client.rs
@@ -72,10 +72,7 @@ mod encryption {
             Mode::Tls => {
                 let config = {
                     let mut config = ClientConfig::new();
-                    config.root_store = match rustls_native_certs::load_native_certs() {
-                        Ok(store) => store,
-                        Err((_, err)) => Err(err)?,
-                    };
+                    config.root_store = rustls_native_certs::load_native_certs().map_err(|(_, err)| err)?;
 
                     Arc::new(config)
                 };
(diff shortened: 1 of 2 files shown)


application-developer-DA:
Ok, I think I sort of understand the purpose of the downgrade as `rustls` of `0.19.1` indeed has the `webpki = 0.21` dependency.

But what about using the `rusttls-native-certs` instead of `webpki-roots`? Why do we want to prefer this over the `webpki-roots`? CC @jxs

src/client.rs
```diff
@@ -72,7 +72,10 @@ mod encryption {
             Mode::Tls => {
                 let config = {
                     let mut config = ClientConfig::new();
-                    config.root_store.add_server_trust_anchors(&webpki_roots::TLS_SERVER_ROOTS);
+                    config.root_store = match rustls_native_certs::load_native_certs() {
+                        Ok(store) => store,
+                        Err((_, err)) => Err(err)?,
+                    };
```
application-developer-DA, line 78: IMHO it would look a bit prettier if we could write it like this:
```suggestion
                    config.root_store = rustls_native_certs::load_native_certs().map_err(|(_, err)| err)?;
```

yusdacra:
> But what about using the `rusttls-native-certs` instead of `webpki-roots`? Why do we want to prefer this over the `webpki-roots`?

`webpki-roots` uses bundled certificates and does not use system certificates. My main motivation here is that local development using `mkcert` is broken because of this. You can see a [comparison](https://github.com/rustls/rustls-native-certs#should-i-use-this-or-webpki-roots) on `rustls-native-certs` repository too.

src/client.rs
yusdacra: I agree that does look better. Though something I forgot to mention is that `load_native_certs()` returns `Result<RootCertStore, (Option<RootCertStore>, Error)>`, the `Option<RootCertStore>` can include certs that are successfully loaded. Do we want to fallback to that in case of an error and log the error, or do we want to keep the behaviour as-is (don't fallback to partially loaded certs, and return the error)?

application-developer-DA: I think failing in case of an error would be fine. I expect that the `TlsError` returned to the user should be sufficient to understand what caused an error. Perhaps we could think of an improvement while working on https://github.com/snapview/tungstenite-rs/issues/203 which looks like a nice thing to have.

application-developer-DA:
Ok, thanks for the info. It looks sensible for me. @jxs if you're fine with it, we could merge I think (once the tiny code review remark is addressed).

Issue #210, opened 2021-06-26
Bad performance with `input_buffer > 3.0` 


qiujiangkun:
The assembling process takes up most of the time in my IO thread. Why is it so slow? I would like to file an PR to improve it.
![image](https://user-images.githubusercontent.com/33482468/123516411-90731400-d6ce-11eb-9892-2dcacdd3363d.png)

qiujiangkun:
Sadly, the performance of `input_buffer` is unacceptable.
https://github.com/qiujiangkun/input_buffer/blob/master/benches/read_data.rs
```log
Benchmarking throughput/decode: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 11.2s, or reduce sample count to 40.
throughput/decode       time:   [110.31 ms 110.61 ms 110.92 ms]                              
                        thrpt:  [216.37 MiB/s 216.98 MiB/s 217.58 MiB/s]
```
https://github.com/snapview/input_buffer/issues/6#issuecomment-869020368

application-developer-DA:
Note that the provided custom "hack" version is essentially the same as older version of the `input_buffer` and the reason it was changed was that its implementation was unsound, so as maintainer I can't recommend using the "HACK" version right now.

Current state and discussion: https://github.com/snapview/input_buffer/issues/6#issuecomment-870548303

We're going to soon release the new version that is as safe as our current sound implementation but significantly faster.

whyCPPgofast:
I see that input_buffer crate was updated and the bottleneck has been fixed. [https://github.com/snapview/input_buffer/pull/9](url)
Is it as simple as bumping the input_buffer version to 0.5.0 in tungstenite or is there more work to do? Thank you for the great library.

application-developer-DA:
You're right, it was just about bumping the version of the dependency and releasing a new version. I decided to postpone it a bit, so that we can merge and release even better version that performs [even faster on average](https://github.com/snapview/tungstenite-rs/pull/214). I expect us delivering a new version tomorrow if the PR gets reviewed today.

whyCPPgofast:
I see you updated the input_buffer crate and added benchmarks. Mind bumping the version of tungstenite-rs ? Thank you for the great work!

PR #214, opened 2021-07-05
Faster implementation of `InputBuffer`

application-developer-DA:
Fixes https://github.com/snapview/tungstenite-rs/issues/210.

Please [check the discussion and decision](https://github.com/snapview/input_buffer/issues/6#issuecomment-870548303) if you don't know the context as well as the [latest PR in `input_buffer`](https://github.com/snapview/input_buffer/pull/9).

It seems like the new simplified implementation has only significant difference on x86_64 architecture while the changes on Armv8.4-A A64 (Apple M1) are not that significant as both versions perform really well. On ARM the new version even behaves a bit slower (but not significantly slower), however the difference on Intel (x86_64) architecture is quite significant.

```
===== M1 Mac (Armv8.4-A A64) =====

buffers/InputBuffer     time:   [543.66 us 549.32 us 555.54 us]
                        thrpt:  [7.0315 GiB/s 7.1110 GiB/s 7.1851 GiB/s]

buffers/ReadBuffer      time:   [552.16 us 565.83 us 581.26 us]
                        thrpt:  [6.7203 GiB/s 6.9036 GiB/s 7.0745 GiB/s]


===== Intel (x86_64) =====

buffers/InputBuffer     time:   [4.7392 ms 4.7601 ms 4.7817 ms]
                        thrpt:  [836.52 MiB/s 840.31 MiB/s 844.03 MiB/s]

buffers/ReadBuffer      time:   [726.86 us 730.72 us 734.81 us]
                        thrpt:  [5.3160 GiB/s 5.3457 GiB/s 5.3741 GiB/s]
```

So it seems like migrating to the newer `ReadBuffer` makes sense as the 'average performance gain' is quite significant (and I believe most users are still using x86_64).

⚠️ Notes:
1. The PR also removes a couple of errors that could have been triggered by `tungstenite-rs`. These were removed as they could only occur if the incoming message does not fit into the hardcoded `u32::max_value()`. Currently we have [`max_frame_size` and `max_message_size`](https://docs.rs/tungstenite/0.13.0/tungstenite/protocol/struct.WebSocketConfig.html), but we don't use them to cap the size of the input/read buffer (instead we compare the values once we read something). This in theory is not a problem when using `tungstenite-rs` with non-blocking sockets (like in case of `tokio-tungstenite`), though it could pose the issue for those ones who use blocking sockets. In other words, the previous reading with `u32::max_value()` did not really add any sort of "additional protection", so I removed them all together to not complicate the `ReadBuffer` implementation.
2. There is no automation to run `criterion` benchmarks with our CI. That's fine as `criterion` is not meant to be used like this and may give many false positives and false negatives when used like this. In order to automate benchmarking with CI we have to use a tool like [lai](https://github.com/bheisler/iai) that was written to be used for such purposes (it works differently than `criterion`), but I think it's better to make it as part of a separate PR.

P.S.: Thanks @qiujiangkun for bringing this topic and performance testing of `tungstenite-rs` 👍

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ byteorder = "1.3.2"
 bytes = "1.0"
 http = "0.2"
 httparse = "1.3.4"
-input_buffer = "0.4.0"
 log = "0.4.8"
 rand = "0.8.0"
 sha-1 = "0.9"
@@ -53,5 +52,6 @@ optional = true
 version = "0.5.0"
 
 [dev-dependencies]
+input_buffer = "0.5.0"
 env_logger = "0.8.1"
 net2 = "0.2.33"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,6 +52,12 @@ optional = true
 version = "0.5.0"
 
 [dev-dependencies]
-input_buffer = "0.5.0"
+criterion = "0.3.4"
 env_logger = "0.8.1"
+input_buffer = "0.5.0"
 net2 = "0.2.33"
+rand = "0.8.4"
+
+[[bench]]
+name = "buffer"
+harness = false
--- a/benches/buffer.rs
+++ b/benches/buffer.rs
@@ -1,0 +1,36 @@
+use std::io::{Cursor, Read};
+
+use criterion::*;
+use input_buffer::InputBuffer;
+use tungstenite::buffer::ReadBuffer;
+
+const CHUNK_SIZE: usize = 4096;
+
+#[inline]
+fn current_input_buffer(mut stream: impl Read) {
+    let mut buffer = InputBuffer::with_capacity(CHUNK_SIZE);
+    while buffer.read_from(&mut stream).unwrap() != 0 {}
+}
+
+#[inline]
+fn fast_input_buffer(mut stream: impl Read) {
+    let mut buffer = ReadBuffer::<CHUNK_SIZE>::new();
+    while buffer.read_from(&mut stream).unwrap() != 0 {}
+}
+
+fn benchmark(c: &mut Criterion) {
+    const STREAM_SIZE: usize = 1024 * 1024 * 4;
+    let data: Vec<u8> = (0..STREAM_SIZE).map(|_| rand::random()).collect();
+    let stream = Cursor::new(data);
+
+    let mut group = c.benchmark_group("buffers");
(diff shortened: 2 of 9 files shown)


agalakhov:
Is there any size limit now? Is it possible to DoS by memory overfull?

application-developer-DA:
> Is there any size limit now? Is it possible to DoS by memory overfull?

Do you mean those points that I mentioned in (1)? If so, then I think there is no problem with it since the previous limit that we had on the `InputBuffer` was a hardcoded value of `usize::max_value()` which I believe could not even be triggered in any normal circumstances because of [`max_frame_size` and `max_message_size`](https://docs.rs/tungstenite/0.13.0/tungstenite/protocol/struct.WebSocketConfig.html) that were checked (and still checked) during the message assembly. Since we read chunks, the only way to trigger the error from the `InputBuffer` was user setting either of `max_frame_size` or `max_message_size` in the `WebSocketConfig` to the `Some(usize::MAX)`. Another case when the limit could hit was the initial handshake. I don't think that the new implementation is worse than **previous** implementation in those particular cases since even if we get to the `usize::MAX`, the size of the underlying `Vec` will be at its maximum value (`len` is `usize`), pushing to such `Vec` would result in panic. But the **old implementation** did not handle it better since it tried to return an error by summing up and checking the `usize` values without taking the overflow into an account. Adding `usize` with some `reserve` would lead to the panic in the old implementation as well.

In other words, the new drop-in replacement is probably not perfect (since there is a way to get the panic), but it is at least not worse than the previous implementation. Please correct me if I'm wrong or missed something.

However while replying to your comment and checking the unit tests (trying to extend them) I noticed another thing: the new implementation had an issue that it did not clean up the data that is read by the cursor (which normally is not required anymore), so I pushed a new commit that implements a behavior similar to the previous `remove_garbage` stage in the `InputBuffer` and extended the unit test to account for that.

With that step implemented the new `ReadBuffer` seems to still be faster.

```
buffers/InputBuffer     time:   [3.1572 ms 3.1728 ms 3.1904 ms]                                 
                        thrpt:  [1.2244 GiB/s 1.2312 GiB/s 1.2373 GiB/s]

buffers/ReadBuffer      time:   [1.2585 ms 1.2659 ms 1.2737 ms]                                
                        thrpt:  [3.0668 GiB/s 3.0857 GiB/s 3.1038 GiB/s]
```

jxs (approved):

Issue #215, opened 2021-07-09
Add feature flag to use webpki-roots instead of rustls-native-certs


lucacasonato:
This was changed in the last release.

dnaka91:
That would be great to have back in, especially as `tokio-tungstenite` still uses `webpki-roots` so both will be compiled even though only one is used and unnecessarily increasing the build time.

Most importantly though, with `webpki-roots` I was able to build Docker images that use the special `scratch` image as base, meaning it will have the bare minimum included and there will be no native certs. So my project will crash on startup due to missing root certs.

I'd be happy to prepare the changes for this, if there is the interest to make a new release soon. Maybe with the feature flags changed to (idea stolen from `reqwest`):
- `rustls-tls-webpki-roots`
- `rustls-tls-native-roots`
- `rustls-tls-manual-roots` (this one probably not for now, but if needed could help to allow providing custom root certificates).

Also, would be interesting to know what's the motivation for this change in the first place. All I could gather from #205 was that `rustls-native-certs` is helpful for local development and if local system configuration is preferred. At least in my case that's not what I need and it would have been great to just handle it through feature flags instead of fully switching over.

application-developer-DA:
> That would be great to have back in, especially as tokio-tungstenite still uses webpki-roots so both will be compiled even though only one is used and unnecessarily increasing the build time.

Ah, you're right. That would be indeed an unnecessary dependency.


> Also, would be interesting to know what's the motivation for this change in the first place. All I could gather from #205 was that `rustls-native-certs` is helpful for local development and if local system configuration is preferred. At least in my case that's not what I need and it would have been great to just handle it through feature flags instead of fully switching over.

Yeah, I think that `rustls-native-certs` should ideally be an option, allowing the user to choose between `webpki-roots` and `rustls-native-certs`. Both [seem to have some pros and cons](https://github.com/rustls/rustls-native-certs#should-i-use-this-or-webpki-roots).

> I'd be happy to prepare the changes for this, if there is the interest to make a new release soon. Maybe with the feature flags changed to (idea stolen from `reqwest`):

This actually sounds good!

dnaka91:
@application-developer-DA I opened a PR with the required changes.
I created it as draft s I still have a few open questions and would like to resolve them first before I consider the PR ready for review (code wise).

dnaka91:
With #217 merged this can be closed maybe?

application-developer-DA:
Oh yeah, thanks for reminding!

PR #217, opened 2021-07-12
Enable root cert providers through feature flags

dnaka91:
Add feature flags for different providers of root certificates as discussed in #215. As mentioned in the issue `rustls` still uses the `0.21` version of `webpki`, that's why the `0.22` version wasn't used here. There's already a `0.20.0-beta2` of `rustls` so after some waiting that issue might be solved with an upgrade of rustls. As that new version brings many breaking changes, I didn't upgrade to the beta version for now.

## Currently open questions

- [x] This change will break implementations for most users as this will not change the API but with only the `rustls-tls` feature enabled there will be no root certs available at runtime. Show we maybe make the new `rustls-tls-*` flags independent of the `rustls-tls` flag and turn `rustls-tls` into an alias for `rustls-tls-native-roots`?
- [x] I believe the `Mode` enum could be an interesting choice to allow for custom certs to be passed in as optional values. That would mean to turn `Mode::Tls` into something like `Mode::Tls { root_certs: Option<TLSServerTrustAnchors> }`, but this must be further investigated to make it compatible for situations where `native-tls` is used instead. Any thoughts?
- [x] The current implementation allows for both new feature flags to be present and results in both native certs as well as embedded certs to be loaded. This should be the preferred way to make everyone happy I hope?
  - [x] Have to test how the native roots behave when there are none available (currently hoping that it won't fail and just returns an empty root store).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+# 0.15.0 (unreleased)
+
+- Allow selecting the method of loading root certificates if `rustls` is used as TLS implementation.
+  - Two new feature flags `rustls-tls-native-roots` and `rustls-tls-webpki-roots` have been added
+    that activate the respective method to load certificates.
+  - The `rustls-tls` flag was removed to raise awareness of this change. Otherwise, compilation
+    would have continue to work and potential errors (due to different or missing certificates)
+    only occurred at runtime.
+  - The new feature flags are additive. If both are enabled, both methods will be used to add
+    certificates to the TLS configuration.
+
 # 0.14.0
 
 - Use `rustls-native-certs` instead of `webpki-root` when `rustls-tls` feature is enabled.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,8 @@ all-features = true
 default = []
 native-tls = ["native-tls-crate"]
 native-tls-vendored = ["native-tls", "native-tls-crate/vendored"]
-rustls-tls = ["rustls", "webpki"]
-rustls-tls-native-roots = ["rustls-tls", "rustls-native-certs"]
-rustls-tls-webpki-roots = ["rustls-tls", "webpki-roots"]
+rustls-tls-native-roots = ["rustls", "webpki", "rustls-native-certs"]
+rustls-tls-webpki-roots = ["rustls", "webpki", "webpki-roots"]
 
 [dependencies]
 base64 = "0.13.0"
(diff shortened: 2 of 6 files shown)


application-developer-DA:
> This change will break implementations for most users as this will not change the API but with only the rustls-tls feature enabled there will be no root certs available at runtime. Show we maybe make the new rustls-tls-* flags independent of the rustls-tls flag and turn rustls-tls into an alias for rustls-tls-native-roots?

From the point of view of a seamless update of the version this seems to be a tempting solution, but I'm not sure if it's worth it in a long run as some may found it confusing that we have three `rustls` features, where one of the features (`rustls-tls`) is just an alias for another (`rustls-tls-native-roots`), i.e. it's likely that in future someone will come up with the suggestion or a PR to remove the alias, so the breaking change will anyway be required at some point. Because of that my personal preference would be to just have `rustls-tls-native-roots` and `rustls-tls-webpki-roots` without having an additional `rustls-tls` alias. We will increase the version to signify a breaking change and add a note to the [changelog](https://github.com/snapview/tungstenite-rs/blob/master/CHANGELOG.md). IMHO it would be the 'cleanest' approach (though it would require the user of the library to change the feature when updating the version of `tungstenite-rs`, but I think updating the version is anyway involves the `Cargo.toml`, so it could be a one-liner + it raises the awareness of the user about which certs are going to be included).


> I believe the Mode enum could be an interesting choice to allow for custom certs to be passed in as optional values. That would mean to turn Mode::Tls into something like Mode::Tls { root_certs: Option<TLSServerTrustAnchors> }, but this must be further investigated to make it compatible for situations where native-tls is used instead. Any thoughts?

I have not yet thought through this idea, but it sounds good. Do I get it right that doing so is pretty much a solution for https://github.com/snapview/tungstenite-rs/issues/203?

> The current implementation allows for both new feature flags to be present and results in both native certs as well as embedded certs to be loaded. This should be the preferred way to make everyone happy I hope?

Ah, you mean, having both `webpki` and native/system roots? Yeah, this might be a good idea in terms of 'compatibility', but I'm not sure if it would be desirable in most cases, i.e. if someone uses `webpki-roots`, then I could imagine that the intent is to explicitly **not** use the native certs and the vice versa, although explicitly telling to use one or another remains a possibility (i.e. the user is free to choose whether to use only `webpki-roots`, only `native-certs` or both). So yeah, maybe this idea is not bad at all 🙂

dnaka91:
I removed the `rustls-tls` feature flag now and tested all combination of features with `cargo-hack`.
Added an entry to the changelog as well 🙇

application-developer-DA (approved):

dnaka91:
Okay, tested that it works in a scratch docker image with both feature flags enabled.
Found a silly but important bug on the way.

Should be good to go now 🙇‍♂️

jxs (approved):
LGTM! thanks!

dnaka91:
Thanks for the quick review and merge 👍

PR #226, opened 2021-07-23
Add a connector to configure TLS config

### LINKED ISSUE #185: Add a connector like in tokio-tungestenite
While I implemented https://github.com/snapview/tokio-tungstenite/pull/147 I really liked the extra `*_tls*` functions that allow to pass a TLS connector. Not just because it allows to configure the connector details, but because it gives the chance to explicitly tell the crate what TLS implementation to use.

I thought that would be a nice addition for this crate as well to get the same benefits as these functions are currently not available here. That would mean to add `client_tls` and `client_tls_with_config` functions, a `Connector` and replacing `AutoStream` with `MaybeTlsStream`.

Just wanted to ask before I start implementing it and would like to hear opinion of the snapview team and community.

### LINKED ISSUE #203: Remove hard webpki-roots dependency by allowing the user to supply a Rustls `ClientConfig` when using Rustls

dnaka91:
This is mostly a backport of what was done in `tokio-tungstenite` by adding a connector. It allows to pick which TLS variant to use and further change the used configuration. That feature was initially discussed in #185. In addition it should solve the request described in #203.

Fixes #185, fixes #203.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,9 @@ all-features = true
 default = []
 native-tls = ["native-tls-crate"]
 native-tls-vendored = ["native-tls", "native-tls-crate/vendored"]
-rustls-tls-native-roots = ["rustls", "webpki", "rustls-native-certs"]
-rustls-tls-webpki-roots = ["rustls", "webpki", "webpki-roots"]
+rustls-tls-native-roots = ["__rustls-tls", "rustls-native-certs"]
+rustls-tls-webpki-roots = ["__rustls-tls", "webpki-roots"]
+__rustls-tls = ["rustls", "webpki"]
 
 [dependencies]
 base64 = "0.13.0"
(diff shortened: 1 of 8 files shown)


application-developer-DA (approved):
Thanks, looks good to me!

jxs (approved):
great work, thanks!

PR #227, opened 2021-07-25
fix: write some header twice

LYF1999:
if specify `host` in headers, will write `host` twice for `req`

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -3,7 +3,7 @@
 use std::{io, result, str, string};
 
 use crate::protocol::{frame::coding::Data, Message};
-use http::Response;
+use http::{header::HeaderName, Response};
 use thiserror::Error;
 
 /// Result type of all Tungstenite library calls.
@@ -138,7 +138,7 @@ pub enum CapacityError {
 }
 
 /// Indicates the specific type/cause of a protocol error.
-#[derive(Error, Debug, PartialEq, Eq, Clone, Copy)]
+#[derive(Error, Debug, PartialEq, Eq, Clone)]
 pub enum ProtocolError {
     /// Use of the wrong HTTP method (the WebSocket protocol requires the GET method be used).
     #[error("Unsupported HTTP method used - only GET is allowed")]
@@ -167,6 +167,10 @@ pub enum ProtocolError {
     /// Custom responses must be unsuccessful.
     #[error("Custom response must not be successful")]
     CustomResponseSuccessful,
+    /// Invalid header is passed. This header is formed by the library automatically
+    /// and must not be overwritten by the user.
+    #[error("Not allowed to pass overwrite the standard header {0}")]
+    InvalidHeader(HeaderName),
     /// No more data while still performing handshake.
     #[error("Handshake not finished")]
     HandshakeIncomplete,
--- a/src/handshake/client.rs
+++ b/src/handshake/client.rs
@@ -5,7 +5,7 @@ use std::{
     marker::PhantomData,
 };
 
-use http::{HeaderMap, Request as HttpRequest, Response as HttpResponse, StatusCode};
+use http::{header, HeaderMap, Request as HttpRequest, Response as HttpResponse, StatusCode};
 use httparse::Status;
 use log::*;
 
@@ -125,6 +125,14 @@ fn generate_request(request: Request, key: &str) -> Result<Vec<u8>> {
     .unwrap();
 
     for (k, v) in request.headers() {
+        if k == header::CONNECTION
+            || k == header::UPGRADE
+            || k == header::SEC_WEBSOCKET_VERSION
+            || k == header::SEC_WEBSOCKET_KEY
+            || k == header::HOST
+        {
+            return Err(Error::Protocol(ProtocolError::InvalidHeader(k.clone())));
+        }
         let mut k = k.as_str();
         if k == "sec-websocket-protocol" {
             k = "Sec-WebSocket-Protocol";


LYF1999:
should we use the same  strategy for `CONNECTION, UPGRADE, SEC_WEBSOCKET_VERSION, SEC_WEBSOCKET_KEY` like `host`?

src/handshake/client.rs
```diff
@@ -108,6 +108,10 @@ fn generate_request(request: Request, key: &str) -> Result<Vec<u8>> {
         return Err(Error::Url(UrlError::EmptyHostName));
     }
 
+    if let Some(v) = request.headers().get(header::HOST) {
+        host = unsafe { std::str::from_utf8_unchecked(v.as_ref()) };
+    }
+
```
application-developer-DA, line 114: But what's the point of parsing the `host` value if we're going to overwrite it in presence of the custom user header? What I mean is that the same logic could then be rewritten like this:

```rust
if let Some(user_host_header) = requests.headers().get(header::HOST) {
    host = ...
} else if let Some(idx) = authority.find('@') {
   ....
}
```

So that we don't need to have a `mut host` and the code is a bit more safe.

In addition, I don't think introducing an `unsafe` block here would be a good decision as the same logic could be written safely.

Speaking more broadly I'm not sure that this change would make sense in general because of the RFC 6455 that states that:

```
       <...> the URI to
       pass the function MUST be constructed from /host/, /port/,
       /resource name/, and the /secure/ flag using the definition of a
       WebSocket URI as given in Section 3.

<...>

   4.   The request MUST contain a |Host| header field whose value
        contains /host/ plus optionally ":" followed by /port/ (when not
        using the default port).
```

I.e. it does not seem like we're allowed to overwrite the `Host` here.

There is [section 12](https://datatracker.ietf.org/doc/html/rfc6455#section-12) in the RFC but I'm not sure if it's the one that you had in mind to apply here:

```
12.  Using the WebSocket Protocol from Other Specifications

   The /host/, /port/, /resource name/, and /secure/ flag are usually
   obtained from a URI using the steps to parse a WebSocket URI's
   components.  These steps fail if the URI does not specify a
   WebSocket.
```

CC @jxs

src/handshake/client.rs
```diff
@@ -125,6 +129,14 @@ fn generate_request(request: Request, key: &str) -> Result<Vec<u8>> {
...
 
     for (k, v) in request.headers() {
+        if k == header::CONNECTION
+            || k == header::UPGRADE
+            || k == header::HOST
+            || k == header::SEC_WEBSOCKET_VERSION
+            || k == header::SEC_WEBSOCKET_KEY
+        {
+            continue;
+        }
```
application-developer-DA, line 139: According to the RFC 6455:

```
   An unordered set of header fields comes after the leading line in
   both cases.  The meaning of these header fields is specified in
   Section 4 of this document.  Additional header fields may also be
   present, such as cookies [RFC6265]. 

   In compliance with [RFC2616], header fields in the handshake may be
   sent by the client in any order, so the order in which different
   header fields are received is not significant.
```

So I'm not sure if we're really prohibited to add additional headers even if they duplicate some of the previous headers provided that we always ensure that the unordered set of the required headers is always passed before user headers, so for the mandatory headers the standard values are going to be used, user values may then be used for other purposes (some custom user logic).

But even if we were to prohibit the user to specify some headers, I think we would better returning an error early instead of silently skipping them (may not be the expected behavior for the user).

src/handshake/client.rs
LYF1999: my scene is that we want to  dial to a shared endpoint, then the shared endpoint can dispatch the differently upstream by `host`.

src/handshake/client.rs
LYF1999: good suggestion, it's better to return a error.

src/handshake/client.rs
application-developer-DA: But it does seem to contradict the RFC, doesn't it?

LYF1999: but some http clients, for example hyper allow users to overwrite `host` in headers . maybe it's a de facto standard

https://github.com/hyperium/hyper/blob/6393a0cc031f504dbd50eb4b021bae51e21f6140/src/client/client.rs#L248-L259
```rs
            if self.config.set_host {
                let uri = req.uri().clone();
                req.headers_mut().entry(HOST).or_insert_with(|| {
                    let hostname = uri.host().expect("authority implies host");
                    if let Some(port) = get_non_default_port(&uri) {
                        let s = format!("{}:{}", hostname, port);
                        HeaderValue::from_str(&s)
                    } else {
                        HeaderValue::from_str(hostname)
                    }
                    .expect("uri host is valid header value")
                });

```

application-developer-DA: Hm, that's interesting. I wonder if any other WebSocket libraries allow it. We could theoretically allow it, it's not a big deal to handle it in the code. What bothered me a little bit is that it's effectively against the RFC sections that I mentioned in my comments. Do I miss something?

LYF1999: gorilla websocket client allow it. https://github.com/gorilla/websocket/blob/e8629af678b7fe13f35dff5e197de93b4148a909/client.go#L207-L224.

hm, it looks like RFC doesn't show `Host` can be allowed to be different with original URI. I think it's more likely to be a de facto standard. 
there some discuss
https://github.com/curl/curl/issues/3664

agalakhov:
I think returning a error would be correct in this case.

BTW, I don't see any reason for using `unsafe` for `from_utf8_unchecked` here unless a benchmark shows significant improvement. Chances are high that the safety check will be optimized-out.

LYF1999:
@application-developer-DA removed overwrite `Host` and unsafe code. PTAL

src/error.rs
```diff
@@ -167,6 +167,9 @@ pub enum ProtocolError {
     /// Custom responses must be unsuccessful.
     #[error("Custom response must not be successful")]
     CustomResponseSuccessful,
+    /// Duplicated header
+    #[error("Duplicated header {0}")]
+    DuplicatedHeader(HeaderName),
```
application-developer-DA, line 172:
```suggestion
    /// Invalid header is passed. This header is formed by the library automatically
    /// and must not be overwritten by the user.
    #[error("Not allowed to pass overwrite the standard header {0}")]
    InvalidHeader(HeaderName),
```

src/handshake/client.rs
```diff
@@ -125,6 +125,14 @@ fn generate_request(request: Request, key: &str) -> Result<Vec<u8>> {
     .unwrap();
 
     for (k, v) in request.headers() {
+        if k == header::CONNECTION
+            || k == header::UPGRADE
+            || k == header::SEC_WEBSOCKET_VERSION
+            || k == header::SEC_WEBSOCKET_KEY
+            || k == header::HOST
+        {
+            return Err(Error::Protocol(ProtocolError::DuplicatedHeader(k.clone())));
```
application-developer-DA, line 134:
```suggestion
            return Err(Error::Protocol(ProtocolError::InvalidHeader(k.clone())));
```

application-developer-DA:
> @application-developer-DA removed overwrite Host and unsafe code. PTAL

Ok, thanks, but this one does not compile at the moment.

LYF1999:
Thanks, fixed the compile problem and updated the error message.

application-developer-DA (approved):
Ok, thanks.

@jxs , if you're also fine, we could merge I think.

PR #229, opened 2021-08-11
Reduce crate package size

dnaka91:
Reduce the crate package size (the file that is uploaded to the creates.io registry) by only including necessary files. This allows faster downloads for everyone.

The include filters were generated with `cargo-diet` and then further adjusted until `cargo package` worked.﻿


application-developer-DA:
Thanks!

PR #234, opened 2021-09-03
Switch from `.unwrap()` to `match`

sigaloid:
Properly match on message, so client disconnect doesn't trigger a panic

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,17 +8,20 @@ use std::thread::spawn;
 use tungstenite::accept;
 
 /// A WebSocket echo server
-fn main () {
+fn main() {
     let server = TcpListener::bind("127.0.0.1:9001").unwrap();
     for stream in server.incoming() {
-        spawn (move || {
+        spawn(move || {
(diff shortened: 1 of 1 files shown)


sigaloid:
Oops, I think this can just become a single `if let Some()`; but perhaps it's better to leave it if someone wants to handle the error.

application-developer-DA:
Thanks for the change, but I unfortunately can't understand the rationale. The reason we used `unwrap()` is just to merely show how it roughly looks like to read a message with `tungstenite-rs` (we of course never use `unwrap()` in such ways in the real production code or `tungstenite-rs` itself). I don't see a point to use the `match` with the empty `Err` handler in the README file as it does not seem to be any related or the `tungstenite-rs` logic, i.e. it's obvious for any Rust user that if function returns a `Result`, then `unwrap()` will panic in case of the error unless the error is properly handled ;)

In fact I also think using the code this way is quite dangerous (silently ignoring an error means that once we get an error, we enter the infinite `loop` with all the consequences).

PR #242, opened 2021-10-15
support overriding DNS items for connect

clia:


clia:
Well, in our situation we need to connect to a virtual host, customizing the DNS resolve items.
When using the reqwest library, it provides the `resolve` method on the `ClientBuilder`, to override DNS resolution for specific domains to particular IP addresses.
I would like the WebSocket library to provide a similar method to do this, since WebSocket is a long-connection protocol, it's sometimes useful to set up a virtual host.

application-developer-DA:
> Well, in our situation we need to connect to a virtual host, customizing the DNS resolve items.

But would it not work by connecting to this host and passing the connection (stream) to the `tungstenite-rs` via `client()`?

> When using the reqwest library, it provides the resolve method on the ClientBuilder, to override DNS resolution for specific domains to particular IP addresses.

That's partially because `reqwest` is a high level wrapper that does not expose lower-level means to just "run http on this S" where `S: Read + Write` (or `AsyncRead + `AsyncWrite` in terms of Tokio). `tungstenite-rs` on other means is all about WebSocket support and it provides a function that allows to run the WebSockets on a given stream. `connect` is just a feature to spare some boilerplate for the very common use case of just connecting to the URL and exchanging messages with it. Unfortunately we can't make `connect` include **all possible** ways to customize the connection procedure: it would produce quite cluttered code with too many [non-obvious] "responsibilities" of the library and confuse users who may be not familiar with a given "extension".

We've also had some internal discussion about this change and I think we'll unfortunately to close this PR as it seems like it goes against the logic/idea described above. However, it's still possible to implement the behavior you want by connecting to the end-point in the user code and passing the resulted stream to the `tungstenite-rs` for WebSockets. Hope this helps!

But in case you have a strong opinion on that or some really good rationale - let us know :slightly_smiling_face:

PR #250, opened 2021-11-02
feature: Raw frame message

Voronar:
According to https://datatracker.ietf.org/doc/html/rfc6455#section-5.4:
```
The primary purpose of fragmentation is to allow sending a message
that is of unknown size when the message is started without having to
buffer that message.  If messages couldn't be fragmented, then an
endpoint would have to buffer the entire message so its length could
be counted before the first byte is sent.  With fragmentation, a
server or intermediary may choose a reasonable size buffer and, when
the buffer is full, write a fragment to the network.

A secondary use-case for fragmentation is for multiplexing, where it
is not desirable for a large message on one logical channel to
monopolize the output channel, so the multiplexing needs to be free
to split the message into smaller fragments to better share the
output channel.
```

PR makes possible to implement some sort of fragmentation outside of library code (ex. https://github.com/Voronar/tokio-tungstenite/blob/message_raw_frame_example/examples/fragment_server.rs#L165).


application-developer-DA:
I quickly went through the suggested changes, the idea (possibilities to have multiplexing etc) is interesting, but we would need to check the implications. The quick look suggests that the `Message::Frame` is something that we can write (allow user to send them), however an attempt to read a message using the regular method will never return `Message::Frame` in its current implementation, right?

We would also need to check if we really comply with all requirements from the section 5.4 in RFC (`The following rules apply to fragmentation:`) to make sure that our logic would remain sound. Though maybe it already is.

Voronar:
Current implementation is the most simple (minimal affection on internal architecture, but new variant has to check in user code) solution that allows passing raw frames (all tests has passed).
And yes, it is relevant only to message writing, reading has no effect and shouldn't I think, because current reading mechanism already correctly parses fragmented messages.

application-developer-DA:
Sounds good!

Yeah, I've also thought on how to make it more elegant. One of the first thoughts were to use different types for incoming / outgoing messages, but I think that although it would look better from the perspective of strong typing system here, the overall solution would be even less elegant due to different types being used (potential issues using it with `split()` / `StreamExt` etc due to different data types). Or the one could use one type `enum` with 2 variants and return this type, but that's even more sketchier approach I think.

Perhaps your solution is indeed the least intrusive and more elegant than other alternatives.

src/protocol/message.rs
```diff
@@ -252,6 +256,7 @@ impl Message {
             }
             Message::Close(None) => Ok(String::new()),
             Message::Close(Some(frame)) => Ok(frame.reason.into_owned()),
+            Message::Frame(frame) => frame.into_string().map_err(Error::from),
```
application-developer-DA, line 259: I btw wonder if `?` would work here (and at line 273 as well).

Voronar: Done.

Voronar:
Ready for merging.

src/protocol/message.rs
```diff
@@ -172,6 +172,8 @@ pub enum Message {
     Pong(Vec<u8>),
     /// A close message with the optional close frame.
     Close(Option<CloseFrame<'static>>),
+    /// Raw frame
```
application-developer-DA, line 175:
```suggestion
    /// Raw frame. Note, that you're not going to get this value while reading the message.
    /// This is meant for the sending part in case sending of the raw frames is desired.
    /// See the [discussion](https://github.com/snapview/tungstenite-rs/pull/250) for more details.
```

Voronar: Thanks. Done.

application-developer-DA:
Alright, thanks, let's give it a try :)

PR #254, opened 2022-01-02
Adds two new public functions to `client.rs`

tlietz:
The two new functions in `client.rs` are `connect_nonblocking` and `connect_nonblocking_with_config`. They are similar to their `connect` counterparts, except they set the inner `TcpStream` of the WebSocket to be non-blocking.

This could be useful for anybody who would like to get a quick non-blocking WebSocket set up for rapid prototyping.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -119,6 +119,48 @@ pub fn connect<Req: IntoClientRequest>(
     connect_with_config(request, None, 3)
 }
 
+/// Connect to the given WebSocket in non-blocking mode.
+///
+/// Uses a websocket configuration passed as an argument to the function. Calling it with `None` is
+/// equal to calling the `connect_nonblocking()` function.
+///
+/// The URL may be either ws:// or wss://.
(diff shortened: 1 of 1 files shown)


application-developer-DA:
Thanks for the PR, but unfortunately I think that it won't quite fit to the crate. Let me elaborate on it a bit: the only reason we initially introduced `connect` was that those ones who just want a very simple WebSocket client for typical small apps that are not really meant for production usage (or some serious networking) can utilize the `connect` and quickly achieve the desired result. It's just a helper for a quite typical use case (back then `async` / `await` did not exist). We already provide quite a few `connect` functions and I think extending the crate by also including 2 functions for an [arguably] seldom case just to spare a couple lines of code is not worth it. I believe that for any real use cases with non-blocking sockets (whether it's prototyping in `async` environment or starting the development of something that is going to hit production rather soon) `tokio-tungstenite` could be used (and it has the desired behavior "by default" due to being `async` / future-aware).

Issue #255, opened 2022-01-14
Allow the user to overwrite the `Host` header supplied in the `Request`


nekolab:
Hi all 

I know this had been discussed in #227  but I'd really want to re-discuss about the decision.

Currently I'm writing a gateway to do reverse proxying on websocket channel, something like Nginx.
As you may know, in this scenario, the common pattern will look like this:
```

 ┌─────────────────┐  ┌──────────────┐
 │     Gateway     │  │     WWW      │
 │     foo.com     │  │127.0.0.1:1010│
 │    0.0.0.0:80   │  └──────────────┘
 │ /               │
 │ 127.0.0.1:1010  │  ┌──────────────┐
 │                 │  │  API Server  │
 │ /api            │  │127.0.0.1:1020│
 │ 127.0.0.1:1020  │  └──────────────┘
 │                 │
 │ /websocket      │  ┌──────────────┐
 │ 127.0.0.1:1030  │  │  Messaging   │
 │                 │  │127.0.0.1:1030│
 └─────────────────┘  └──────────────┘

```
The gateway will act as a relay, forward the http request and websocket message to the real server.

Here comes the question:
The real uri used to connect is `ws://127.0.0.1:1030/websocket` but the service wants to make sure the
request is from `foo.com` only, so it need to check the `Host` is `foo.com`. Since we forbid modify `host`
header by passing request object, there's no way to make a correct request.

I've read #227 carefully, but a bit confused.

First, let's take a look on RFC
```
12.  Using the WebSocket Protocol from Other Specifications

   The /host/, /port/, /resource name/, and /secure/ flag are usually
   obtained from a URI using the steps to parse a WebSocket URI's
   components.  These steps fail if the URI does not specify a
   WebSocket.
```
It says `/host/ are usually obtained from a URI`, but in some unusual case (like mine), the `host` need
and should could be specified by the caller, this behavior is not violate the RFC

Further more, in my opinion, the RFC describes what the correct protocol should be, but not limiting the
implementation's expand power based on correct behavior by default. Some client (which face to public
end user and have high specialized abilities, like a browser) may limiting the ability due to security concern
and product design but I don't think tungstenite-rs is one of them. 

In general, a library or a tool face to developer seems should provide a correct default value, and allow
end user modify it, at their own risk. it's not a de facto standard violates the RFC, but only provides more
ability. Please think about the http library, we can use it constructs dozens of request violates the RFC but
[...]

PR #258, opened 2022-01-31
Allow user to overwrite the |host| header while generating the request

nekolab:
As we discussed in #255, it's possible for us to provide
some way to sepcifiy the generated request's |host| header.

Now when the incoming request stuct contains a |host| header,
we will use it's value as outgoing |host| header's value.

If there's no |host| header specified, we still use the authority
part of uri as |host| header.

This behavior is complied with RFC 6455 and close issue #255

### DIFF
--- a/src/handshake/client.rs
+++ b/src/handshake/client.rs
@@ -96,17 +96,22 @@ impl<S: Read + Write> HandshakeRole for ClientHandshake<S> {
 fn generate_request(request: Request, key: &str) -> Result<Vec<u8>> {
     let mut req = Vec::new();
     let uri = request.uri();
+    let headers = request.headers();
 
-    let authority = uri.authority().ok_or(Error::Url(UrlError::NoHostName))?.as_str();
-    let host = if let Some(idx) = authority.find('@') {
-        // handle possible name:password@
-        authority.split_at(idx + 1).1
+    let host = if headers.contains_key(header::HOST) {
+        headers.get(header::HOST).unwrap().to_str()?
     } else {
-        authority
+        let authority = uri.authority().ok_or(Error::Url(UrlError::NoHostName))?.as_str();
+        if authority.is_empty() {
+            return Err(Error::Url(UrlError::EmptyHostName));
+        }
+        if let Some(idx) = authority.find('@') {
+            // handle possible name:password@
+            authority.split_at(idx + 1).1
+        } else {
+            authority
+        }
     };
-    if authority.is_empty() {
-        return Err(Error::Url(UrlError::EmptyHostName));
-    }
 
     write!(
         req,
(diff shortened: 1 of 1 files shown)


nekolab:
Hi @application-developer-DA , would you like take a look on this? Thanks

src/handshake/client.rs
```diff
@@ -96,17 +96,22 @@ impl<S: Read + Write> HandshakeRole for ClientHandshake<S> {
...
     let mut req = Vec::new();
     let uri = request.uri();
+    let headers = request.headers();
 
-    let authority = uri.authority().ok_or(Error::Url(UrlError::NoHostName))?.as_str();
-    let host = if let Some(idx) = authority.find('@') {
-        // handle possible name:password@
-        authority.split_at(idx + 1).1
+    let host = if headers.contains_key(header::HOST) {
+        headers.get(header::HOST).unwrap().to_str()?
```
application-developer-DA, line 102: I think this can be rewritten as:

```rust
if let Some(host) = headers.get(header::HOST) {
}
```

Then we don't need to `unwrap()` additionally.

application-developer-DA (changes requested):

application-developer-DA:
I've given it a bit of thought and I think I found a solution that will be flexible enough (would allow the user to specify any headers) while not compromising the simplicity of the `"wss://something.com/ws".into_client_request()`. I'll try to come up with a PR that supersedes the changes within a week.

Basically, the idea is that we accept `impl IntoClientRequest`, the trait is implemented for `http::Request` and for common types like `String`, `&str`. Then we generate the request inside the `ClientHandshake` by creating a new request, taking some parts from the `http::Request` that we get as an input.

I have in mind to change the behavior as follows: change the implementation of the `IntoClientRequest` trait for different types so that:
- The implementation for types like `&str` and `String` does generate the request (as we did it before), i.e. we construct a new request, make sure that we fill up all the fields as specified in the standard, etc.
- The implementation for types like `http::Request` will be changed to pretty much no-op in a sense that if we really do get the request constructed by the user as `http::Request` type, then it means that it was constructed manually and that we probably need to forward it as it **instead of** **selectively** taking the parts of this request and turning it into a proper request as we do now.

I think this will be a logical behavior that is sound and logical: e.g. most users that just use `connect()` will not observe any changes and for them, it will work as expected, while the users who have some custom logic (i.e. construct their request for one or another reason) have this request bypassed (without us altering or changing something in it).

This will of course be a breaking change (semantically) and we'll update the documentation for the function and the changelog accordingly.

In addition, to allow to stick to the standard, what we could do on our side is to make sure that the request does contain all the headers as specified in RFC in a proper format and return an error immediately to the user, e.g. fail early instead of passing the faulty request and waiting for it to fail because of server not accepting it (e.g. we could check that all those `WebSocket*` headers are present, etc).

So far it's my best attempt to come up with a solution for this problem :)

src/handshake/client.rs
nekolab: But the `host` will not live long enough for `write!` use, it's only available in `if` block

nekolab:
> I've given it a bit of thought and I think I found a solution that will be flexible enough

Great!
That's basically what I want to do when issuing #255, allows user use their own request.
It's happy to see we've finally reach the agreement and can't wait to see the change lands.

src/handshake/client.rs
application-developer-DA: Won't it? I think its lifetime should be identical in both cases (as it is the same as `headers`), but I might be wrong.

application-developer-DA:
Ok, I've overhauled the request generation and hopefully, the new behavior makes more sense (it will cover the issue with `Host` as well as with all other headers).

Please take a look: https://github.com/snapview/tungstenite-rs/pull/259

application-developer-DA:
Closing as it was superseded by https://github.com/snapview/tungstenite-rs/issues/255

Issue #263, opened 2022-02-15
Capturing identifiers in format strings break compatibility with rustc < 1.58


mfelsche:
This morning, upon upgrading to `0.17` our CI failed to compile with the following error log:

```
   Compiling tungstenite v0.17.0
error: there is no argument named `header`
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/tungstenite-0.17.0/src/handshake/client.rs:134:22
    |
134 |         write!(req, "{header}: {value}\r\n", value = value.to_str()?).unwrap();
    |                      ^^^^^^^^
 ```
 
We are still on rust `1.56.1` on our main branch, this is why this failed. We will switch soon, so this is not a biggie for us, but it would be nice, to have changes like this be guarded with a `MSRV` in your `Cargo.toml`. Should I add a PR for switching back to the old and painful way of handling format strings for increased compatibility?

application-developer-DA:
Indeed. I think my preference would be us updating the `Cargo.toml` to specify the MSRV.

At the same time, I understand that other products may rely on an older version of the `rustc` for one reason or another. If needed, we could release a patch version (create a branch from the latest tagged release) that is backward compatible with the older compiler version.

However, if it's not a biggie for you (I suppose it's not based on your message 🙂), then I would prefer just mentioning the MSRV in our `Cargo.toml`.

mfelsche:
That would be totally fine with me. Lets hope not too many downstream users are stuck with older rustc versions. If i may note that this is a very subtle way to bring people to the newest Rustc. ;-)

application-developer-DA:
Alright 🙂 New version published.

mfelsche:
Thank you so much!

PR #266, opened 2022-02-21
Downgrade MSRV to 1.56

teozkr:
This is currently blocking us for [upgrading kube-rs](https://github.com/kube-rs/kube-rs/pull/839), since we have some build time regressions on rustc 1.57-1.58.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ documentation = "https://docs.rs/tungstenite/0.17.1"
 repository = "https://github.com/snapview/tungstenite-rs"
 version = "0.17.1"
 edition = "2018"
-rust-version = "1.58"
+rust-version = "1.56"
 include = ["benches/**/*", "src/**/*", "LICENSE-*", "README.md", "CHANGELOG.md"]
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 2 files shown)


application-developer-DA:
Hm, we've recently decided to use the `1.58` when a similar question [has been raised](https://github.com/snapview/tungstenite-rs/issues/263) as we generally do prefer to use the latest Rust compiler version (not only because of the newer features that we seldomly use, but also due to the CVEs). Downgrading for `1.56` is certainly not a big deal though.

NB: As far as I understand, updating to `1.58` is not an option for you due to https://github.com/kube-rs/kube-rs/issues/746? I'm asking since https://github.com/rust-lang/rust/issues/91598 seems to be closed.

teozkr:
Yeah, I'm currently stuck on 1.56 due to https://github.com/kube-rs/kube-rs/issues/746. https://github.com/rust-lang/rust/issues/91598 is closed because the regression has been fixed in rustc 1.59, which is set to be promoted to GA in a few weeks.

> Hm, we've recently decided to use the 1.58 when a similar question https://github.com/snapview/tungstenite-rs/issues/263 as we generally do prefer to use the latest Rust compiler version (not only because of the newer features that we seldomly use, but also due to the CVEs). Downgrading for 1.56 is certainly not a big deal though.

Agreed for the general case, application developers should use the latest rustc unless they have a specific reason not to!

teozkr:
Thanks!

PR #267, opened 2022-02-21
`HeaderName::from_static` requires all-lowercase HTTP2 compatible hea…

sdroege:
…der names

and was passed header names with uppercase characters instead, which
made it panic.

### DIFF
--- a/src/handshake/client.rs
+++ b/src/handshake/client.rs
@@ -114,7 +114,9 @@ fn generate_request(mut request: Request) -> Result<(Vec<u8>, String)> {
         .headers()
         .get(KEY_HEADERNAME)
         .ok_or_else(|| {
-            Error::Protocol(ProtocolError::InvalidHeader(HeaderName::from_static(KEY_HEADERNAME)))
+            Error::Protocol(ProtocolError::InvalidHeader(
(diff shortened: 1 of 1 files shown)


sdroege:
See https://docs.rs/http/latest/http/header/struct.HeaderName.html#method.from_static

sdroege:
Also it's unclear to me why none of the existing tests caught this.

sdroege:
That would be the other option. I can update the PR accordingly, your decision which variant you prefer :)

sdroege:
Well, calling `HeaderName::from_bytes()`.

That's btw called implicitly by the `remove()` and `get()` calls on the strings too and does heap allocations. If that's something you worry about then the list of header names should be stored as actual `HeaderName`s instead of strings statically (e.g. via `once_cell::sync::Lazy`).

sdroege:
Sure, updated accordingly

application-developer-DA:
Thanks! 👍

sdroege:
Can you release a 0.17.2 with this fix in the near future?

application-developer-DA:
Sure! I'll try to publish it today along with the updated `tokio-tungstenite`.

sdroege:
Nice, thanks! I noticed this btw from a user of async-tungstenite. I've released a version of that using tungstenite 0.17 during the weekend.

PR #271, opened 2022-03-15
Update autobahn/expected-results.json for #246

kazk:

PR #272, opened 2022-03-16
Fix autobahn scripts

kazk:
This should fail because of #271.


application-developer-DA:
Oh, how we missed it 😱 

Thanks for the fix! 👍 🚀

application-developer-DA (approved):

PR #287, opened 2022-06-18
fix: uppercase origin header

n3oney:
While HTTP headers should be case-insensitive, unfortunately some servers don't conform to that behavior. Unfortunately, the `http` crate always stores headers in lowercase without any way to prevent that behavior, so if we run into a server that does not conform to spec, we're doomed.
This is a workaround to that issue - it changes the `origin` header into `Origin`.

### DIFF
--- a/src/handshake/client.rs
+++ b/src/handshake/client.rs
@@ -158,6 +158,10 @@ fn generate_request(mut request: Request) -> Result<(Vec<u8>, String)> {
             name = "Sec-WebSocket-Protocol";
         }
 
+        if name == "origin" {
+            name = "Origin";
+        }
+
         writeln!(req, "{}: {}\r", name, v.to_str()?).unwrap();
     }
 


application-developer-DA:
Hm, alright, let's merge it! This is a second header for which we implement such a workaround, perhaps we would need to find a generic solution for it (similar to how `hyper` handled it once) since it's not clear how many other headers will be ignored by the server for the same reason (the server that performs a case-sensitive check of a single header is likely to do the same for all other headers as well).

n3oney:
> Hm, alright, let's merge it! This is a second header for which we implement such a workaround, perhaps we would need to find a generic solution for it (similar to how `hyper` handled it once) since it's not clear how many other headers will be ignored by the server for the same reason (the server that performs a case-sensitive check of a single header is likely to do the same for all other headers as well).

I definitely agree. I think it'd be much better to implement some kind of custom hashmap to handle headers. I know that [isahc](https://crates.io/crates/isahc) has a `title_case_headers` configuration option which as the name suggests turns headers into title case, however that's still very much a workaround, and a full-fledged solution would be appreciated.

Issue #288, opened 2022-07-02
Introduce a method which will retry on WouldBlock


lyonbeckers:
Hi all, I've been attempting to connect using a `mio::net::TcpStream`, using the `client` method for a non-blocking method of connecting. The only way that I've been able to get it to work is to make it so that `handshake()` doesn't return an error when the error is `WouldBlock`. This makes sense to me, given the following from the documentation for `Poll` in mio.

> If operation fails with [WouldBlock](https://doc.rust-lang.org/1.55.0/std/io/error/enum.ErrorKind.html#variant.WouldBlock), then the caller should not treat this as an error, but instead should wait until another readiness event is received.

And

> Once a readiness event is received, the corresponding operation must be performed repeatedly until it returns [WouldBlock](https://doc.rust-lang.org/1.55.0/std/io/error/enum.ErrorKind.html#variant.WouldBlock). 

Which unfortunately for me, my corresponding operation is the `client` call, which simply return an error in this case, and I can't really repeat, since it consumes the stream via move.

application-developer-DA:
Sorry, I did not get what exactly did not work for you. Are you asserting that `client` function cannot be used with `mio`'s  `TcpStream`?

> Which unfortunately for me, my corresponding operation is the client call, which simply return an error in this case, and I can't really repeat, since it consumes the stream via move.

If stream blocks, that's when you get the [`Interrupted` error](https://docs.rs/tungstenite/latest/tungstenite/handshake/enum.HandshakeError.html). Once the socket gets ready, you can repeat it by calling the [`handshake()`](https://docs.rs/tungstenite/latest/tungstenite/handshake/struct.MidHandshake.html#method.handshake) on the corresponding handshake that you got back.

application-developer-DA:
I hope the question was answered 🙂 Let me know if I missed something or if you find some issue with trying to implement it like this (I was quite sure that it would work that way with `mio` since we've been using it in a similar way on our production systems since a couple years).

lyonbeckers:
Hey! Sorry for the late reply, this did absolutely solve the problem for me, thank you so much for the help! It hadn't occurred to me that I could call `handshake()` on the MidHandshake if I unpacked the error

PR #289, opened 2022-07-02
Introduce methods for more permissive client handshakes

lyonbeckers:
Per issue #288, this is a solution that has worked for me to get a proper handshake when using mio. I didn't want to introduce potential breaking changes by making the regular handshake method more permissive, so I just presented an alternate `allow_blocking` path


lyonbeckers:
Closed because it turned out not to be necessary

PR #291, opened 2022-07-26
add support for batching websocket messages

RustUser246:
Adds support for sending multiple messages with a single write call
* Adds a new function for appending messages without trying to write them to the socket immediately. 
* This allows for a significant improvement to throughput and latency by avoiding calling write for each individual message.

### DIFF
--- a/src/protocol/frame/mod.rs
+++ b/src/protocol/frame/mod.rs
@@ -66,13 +66,9 @@ impl<Stream> FrameSocket<Stream>
 where
     Stream: Write,
 {
-    /// Write a frame to stream.
-    ///
-    /// This function guarantees that the frame is queued regardless of any errors.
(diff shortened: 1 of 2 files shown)


agalakhov:
Very good, thanks!

I think the `append_message` name is misleading, maybe `queue_message` and `queue_frame` are better? What do you think?

RustUser246:
Agreed, they're somewhat misleading. I'll rename the functions.

agalakhov:
Applied, thanks!

agalakhov:
Sorry, but your pull request caused **regression** in queue behavior, so I had to revert it.

The cause is, you move all the elements from the queue into the sent buffer. No size limits, no nothing. Then the system tries to send the _whole_ queue in a single system call which easily fails.

In order to combine messages, one has to add size limit for combined messages. Combining must be done immediately before sending, not earlier.

PR #294, opened 2022-08-03
tests: add a regression test for the queue logic

application-developer-DA:
We've recently [had a new PR](https://github.com/snapview/tungstenite-rs/pull/291) that added some useful things but unfortunately caused a regression in the queue logic. The PR was consequently reverted, but one of the key learnings was that we did not have a queue logic covered by a test.

It would be nice if we had a test in place to cover this case. This PR adds such a test. Having this test would allow us to rebuild the desired logic in the correct way while keeping the queue logic sound.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -688,6 +688,64 @@ mod tests {
         }
     }
 
+    struct WouldBlockStreamMoc;
+
+    impl io::Write for WouldBlockStreamMoc {
+        fn write(&mut self, _: &[u8]) -> io::Result<usize> {
+            Err(io::Error::new(io::ErrorKind::WouldBlock, "would block"))
+        }
+        fn flush(&mut self) -> io::Result<()> {
+            Err(io::Error::new(io::ErrorKind::WouldBlock, "would block"))
+        }
+    }
+
+    impl io::Read for WouldBlockStreamMoc {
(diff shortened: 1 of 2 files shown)


src/protocol/mod.rs
```diff
@@ -688,6 +688,64 @@ mod tests {
...
+            assert!(matches!(
+                socket.write_message(message.clone()).unwrap_err(),
+                Error::SendQueueFull(..)
+            ));
+        }
+
+        // The size of the output buffer must not be bigger than the size of that message
+        // that we managed to write to the output buffer at first. Since we could not make
+        // any progress (because of the logic of the moc buffer), the size remains unchanged.
+        if socket.context.frame.output_buffer_len() > message.len() {
```
agalakhov, line 744: This could be less strict. If we're sending lots of small messages, adding some more to the output buffer while waiting could improve performance. I think about introducing a limit (`max_send_at_once`) and using it for combining small messages and/or for splitting large ones.

application-developer-DA: > This could be less strict. If we're sending lots of small messages, adding some more to the output buffer while waiting could improve performance.

If we had a slightly different implementation, it would be likely, agree, but even in such a case it would affect the buffer **capacity**, not the size (length), wouldn't it?

agalakhov: It would affect the amount of messages in the buffer. It won't be equal to the size of the very first message.

PR #298, opened 2022-08-27
Error response on client type returns request body upon non receiving non 101 status code from server

### LINKED ISSUE #199: Read the body of the error response on the client if the server replies with an error
See https://github.com/snapview/tokio-tungstenite/issues/162 for the details.

JosiahBull:
modified `client::Response` type to contain `String` instead of `()` to achieve this.

I ran into this issue yesterday when my automated test suite failed unexpectedly. Notably, fixing this doesn't conflict with RFC 6455, or RFC 2616.

> If the status code received from the server is not 101, the
       client handles the response per HTTP [[RFC2616](https://www.rfc-editor.org/rfc/rfc2616)] procedures.  In
       particular, the client might perform authentication if it
       receives a 401 status code; the server might redirect the client
       using a 3xx status code (but clients are not required to follow
       them), etc.  Otherwise, proceed as follows.

So I've given it a shot. Modified `Response` to wrap a `String` instead of `()` - a significant breaking api change. Would it be better to return a `Vec<u8>` or `Cow`? Or is there some way to do this that doesn't require modifying `Response<()>`?

Keen to hear other suggestions as to avoid the breaking api change. :)

closes #199 and addresses snapview/tokio-tungstenite#162

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -65,7 +65,7 @@ pub enum Error {
     /// HTTP error.
     #[error("HTTP error: {}", .0.status())]
     #[cfg(feature = "handshake")]
-    Http(Response<Option<String>>),
+    Http(Response<Vec<u8>>),
     /// HTTP format error.
     #[error("HTTP format error: {0}")]
     #[cfg(feature = "handshake")]
--- a/src/error.rs
+++ b/src/error.rs
@@ -65,7 +65,7 @@ pub enum Error {
     /// HTTP error.
     #[error("HTTP error: {}", .0.status())]
     #[cfg(feature = "handshake")]
-    Http(Response<Vec<u8>>),
+    Http(Response<Option<Vec<u8>>>),
     /// HTTP format error.
     #[error("HTTP format error: {0}")]
     #[cfg(feature = "handshake")]
--- a/src/handshake/server.rs
+++ b/src/handshake/server.rs
@@ -30,7 +30,7 @@ pub type Request = HttpRequest<()>;
 pub type Response = HttpResponse<()>;
 
 /// Server error response type.
-pub type ErrorResponse = HttpResponse<Vec<u8>>;
+pub type ErrorResponse = HttpResponse<Option<String>>;
 
 fn create_parts<T>(request: &HttpRequest<T>) -> Result<Builder> {
     if request.method() != http::Method::GET {
@@ -265,7 +265,9 @@ impl<S: Read + Write, C: Callback> HandshakeRole for ServerHandshake<S, C> {
                         let mut output = vec![];
                         write_response(&mut output, resp)?;
 
-                        output.extend_from_slice(resp.body());
+                        if let Some(body) = resp.body() {
+                            output.extend_from_slice(body.as_bytes());
+                        }
 
                         ProcessingResult::Continue(HandshakeMachine::start_write(stream, output))
                     }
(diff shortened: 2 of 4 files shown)


application-developer-DA:
Actually, it makes sense and looks quite clean, thanks!

If I remember correctly, the type of the `Response` was `Option<String>` some long time ago, but I don't remember why we changed it 🤔 I think breaking change is fine (we do it from time to time because of the API surface, new features, etc). Regarding the type: Something like `Bytes` or `Vec<u8>` would indeed be a bit more flexible I think. I don't know if it could contain any non-UTF-8 characters in practical cases, but I've noticed that you're performing a lossy conversion into `String`, so maybe having `Vec<u8>` instead would not be a big deal (at least we pass the bytes "as is" and it's up to the user to decide if they need/want to convert it to any other type, including `String`).

What do you think?

JosiahBull:
A `Vec<u8>` would have the benefit of avoiding the extra allocation to a string during an error state too! It appears the server-side of the `tungstenite-rs` crate is designed to return an `Option<String>` as it's error type, though it doesn't ever actually do this.

It can return an `Option<String>` as seen [here](https://github.com/snapview/tungstenite-rs/blob/master/src/handshake/server.rs#L33), but it's prevented due to the call to `create_response` instead of `create_response_with_body` [here](https://github.com/snapview/tungstenite-rs/blob/master/src/handshake/server.rs#L239-L243).

Happy to change the PR to a `Vec<u8>` and leave the decision up to the user, as I agree it's good not to be opinionated about supported encoding (utf8 vs utf16 for example) - but should the scope of the PR expand to include changes to the server side too?

application-developer-DA:
> It can return an Option<String> as seen [here](https://github.com/snapview/tungstenite-rs/blob/master/src/handshake/server.rs#L33), but it's prevented due to the call to create_response instead of create_response_with_body [here](https://github.com/snapview/tungstenite-rs/blob/master/src/handshake/server.rs#L239-L243).

That one is used when the user supplied a callback as far as I can remember. For instance, if the callback has failed, we're setting this error and writing it into the response.

> Happy to change the PR to a Vec<u8> and leave the decision up to the user, as I agree it's good not to be opinionated about supported encoding (utf8 vs utf16 for example) - but should the scope of the PR expand to include changes to the server side too?

I'm ok with small scoped PR 🙂 So even if it improves the behavior on the client side with the change to the `Vec<u8>` that we agree upon, I think it's ok. Further changes could then be worked on in the upcoming PRs if necessary.

JosiahBull:
Now returning `Vec<u8>` instead, note this required changing the Error::Http to contain said type - which resulted in some changes to the serverside api too.

src/error.rs
```diff
@@ -63,7 +63,7 @@ pub enum Error {
     Url(#[from] UrlError),
     /// HTTP error.
     #[error("HTTP error: {}", .0.status())]
-    Http(Response<Option<String>>),
+    Http(Response<Vec<u8>>),
```
application-developer-DA, line 66: Won't it make a bit more sense if we keep it as `Option` but with `Vec<u8>` inside?

This means that the code will be compatible with the previous semantics (if no body is set, it will be `None`), and also it would allow us to differentiate between the "empty body" and "no body".

What do you think?

src/handshake/server.rs
```diff
@@ -30,7 +30,7 @@ pub type Request = HttpRequest<()>;
 pub type Response = HttpResponse<()>;
 
 /// Server error response type.
-pub type ErrorResponse = HttpResponse<Option<String>>;
+pub type ErrorResponse = HttpResponse<Vec<u8>>;
```
application-developer-DA, line 33: I think this one (on the server side) should be kept as `Option<String>`.

Rationale: `ErrorResponse` in this context is also a response that the user returns from their own callback that is supplied.

application-developer-DA (changes requested):
Sorry for the late reply, almost forgot about the PR :)

I've left some comments.

There also seems to be a minor merge conflict after one of the recent PRs was merged.

src/error.rs
JosiahBull: Agreed, will update this. :+1:

JosiahBull: It makes sense to switch the user side to `Option<Vec<u8>>` too, as the `server` and `client` modules currently use a common error type.

JosiahBull:
Rebased onto master to resolve merge conflicts ^^

daniel-abramov (approved):
LGTM.

@agalakhov, if you're ok (if I have not missed anything), let's merge it.

PR #301, opened 2022-08-30
Make handshake dependencies optional.

Icelk:
Second try's the charm


application-developer-DA (approved):

sdroege:
This is a breaking change as code that compiled before with `no-default-features=true` will now not compile anymore if it required the handshake feature. This includes `tokio-tungstenite`, for example.

PR #313, opened 2022-11-06
New API `connect_to_raw_stream` added

ssrlive:
We need a new API that create WebSocket object from an exist socket. likes
```
        let url = "ws://123.45.67.89:1002/my_path";
        let stream = std::net::TcpStream::connect(("123.45.67.89", 1002))?;
        let (mut socket, response) = tungstenite::connect_to_raw_stream(url, stream)?;
        ...
```


agalakhov:
This API already exists.
https://docs.rs/tungstenite/0.17.3/tungstenite/client/fn.client.html

PR #315, opened 2022-11-12
Mark some functions to public

ssrlive:
I'm writing a proxy app that needs to pre-read the data received inside the stream, and then try to parse the data to see if it conforms to the websocket protocol, rather than having tungstenite consume the data. Therefore, some functions of the parsing and encoding engine of the websocket protocol need to be called, and these functions are hidden. So I should make them public.

### DIFF
--- a/src/handshake/machine.rs
+++ b/src/handshake/machine.rs
@@ -1,3 +1,5 @@
+//! handshake machine.
+
 use bytes::Buf;
 use log::*;
 use std::io::{Cursor, Read, Write};
--- a/src/handshake/machine.rs
+++ b/src/handshake/machine.rs
@@ -101,6 +101,7 @@ pub enum RoundResult<Obj, Stream> {
 #[derive(Debug)]
 pub enum StageResult<Obj, Stream> {
     /// Reading round finished.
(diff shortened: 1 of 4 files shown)


src/handshake/machine.rs
```diff
@@ -1,3 +1,5 @@
+//! handshake machine.
```
daniel-abramov, line 1:
```suggestion
//! WebSocket handshake machine.
```

src/handshake/machine.rs
```diff
@@ -101,6 +103,7 @@ pub enum RoundResult<Obj, Stream> {
 #[derive(Debug)]
 pub enum StageResult<Obj, Stream> {
     /// Reading round finished.
+    #[allow(missing_docs)]
```
daniel-abramov, line 106: Hm, why was this required? The `DoneReading` is documented...

daniel-abramov (changes requested):

src/handshake/machine.rs
ssrlive: I don't know. but without it, an error will occure.

PR #316, opened 2022-11-14
`UnexpectedEof` returned after sending and receiving close frame

vilgotf:
After having sent and received a close frame, calling `read_message` or `write_pending` returns `Error::Io` of type `UnexpectedEof` instead of a `Error::ConnectionClosed`. This PR tries to fix this by unconditionally handling `UnexpectedEof` errors as having succesfully read 0 bytes from the stream.

Another way to fix this would be to change `WebSocketState::check_active` to return `Error::ConnectionClosed` for `WebSocketState::CloseAcknowledged`, but I think that modifies `WebSocketState`'s semantics.

### DIFF
--- a/src/protocol/frame/mod.rs
+++ b/src/protocol/frame/mod.rs
@@ -151,7 +151,13 @@ impl FrameCodec {
             }
 
             // Not enough data in buffer.
-            let size = self.in_buffer.read_from(stream)?;
+            let size = self.in_buffer.read_from(stream).or_else(|error| {
+                if error.kind() == IoErrorKind::UnexpectedEof {
+                    Ok(0)
+                } else {
+                    Err(error)
+                }
+            })?;
             if size == 0 {
                 trace!("no frame received");
                 return Ok(None);


agalakhov:
This looks like a hack. I believe it is better to convert `UnexpectedEof` to `ConnectionClosed` explicitly.

daniel-abramov:
I generally would agree with @agalakhov here. The thing is, unconditionally treating `UnexpectedEof` as `ConnectionClosed` doesn't seem to be the nicest idea, after all, normally we don't receive this error. I actually wonder why in your case it's always `UnexpectedEof`, I don't remember observing this error. Is there a specific pattern or a unit test that you could write to trigger the error? - That would allow to investigate it deeper.

vilgotf:
I've observed this in [`twilight-gateway`](https://docs.rs/twilight-gateway), which uses `tokio-tungstenite` wrapped over `rustls` and `tokio::net::TcpSocket`.

daniel-abramov:
Could the problem be on the other side? I.e. them closing the socket earlier than they are supposed to or something like this?

vilgotf:
The error is not in tungstenite!
The following reproduces the error:
```rust
use tokio_stream::StreamExt;
use tokio_tungstenite::{connect_async, tungstenite::Error};
use tracing_subscriber::filter::LevelFilter;

#[tokio::main(flavor = "current_thread")]
async fn main() {
    tracing_subscriber::fmt()
        .with_max_level(LevelFilter::TRACE)
        .init();

    let (mut socket, _) = connect_async("wss://gateway.discord.gg")
        .await
        .expect("Can't connect");

    let hello = socket.next().await.unwrap().unwrap();
    println!("{hello}");
    socket.close(None).await.unwrap();
    loop {
        match socket.next().await {
            Some(Ok(msg)) => println!("{msg}"),
            Some(Err(e)) => panic!("{e}"),
            None => break,
        };
    }
}
```
Whereas using just tungstenite does not error:
```rust
use tracing_subscriber::filter::LevelFilter;
use tungstenite::{connect, Error};

fn main() {
    tracing_subscriber::fmt()
        .with_max_level(LevelFilter::TRACE)
        .init();

    let (mut socket, _) = connect("wss://gateway.discord.gg").expect("Can't connect");

    let hello = socket.read_message().unwrap();
    println!("{hello}");
    socket.close(None).unwrap();
    loop {
        match socket.read_message() {
            Ok(msg) => println!("{msg}"),
            Err(Error::ConnectionClosed) => break,
            Err(e) => panic!("{e}"),
        };
    }
}
```

vilgotf:
I've tracked down the root issue to tokio-rustls.

PR #328, opened 2023-01-07
Add `permessage-deflate` support

### LINKED ISSUE #2: Support for permessage-deflate [✅/awaiting_dependency_merge]
<s>This is a mandatory part of the spec, and is thus needed to be Web-compatible.</s>

kazk:
Supersedes #235.

- Adds optional de/compression logic on top of the existing flow without any _major_ breaking changes.
- `permessage-deflate` is not enabled by default and requires a `deflate` feature.
- `client_max_window_bits` and `server_max_window_bits` are not supported yet because these require `flate2/zlib` which is not pure Rust.
- Passes Autobahn except for 13.3 to 13.6. These require supporting max window bits.
- Correct extension negotiation. The server declines invalid offers, and the client errors on invalid response to terminate the connection.
- `Sec-WebSocket-Extensions` header is parsed by `headers`
    - [ ] Waiting for https://github.com/hyperium/headers/pull/88
- https://github.com/snapview/tokio-tungstenite/pull/251
- Clean `warp` integration: https://github.com/seanmonstar/warp/pull/1016

I might need to adjust some code depending on https://github.com/hyperium/headers/pull/88, but this should be ready for review.

---

Closes #2.

### DIFF
--- a/src/extensions/compression/deflate.rs
+++ b/src/extensions/compression/deflate.rs
@@ -1,0 +1,442 @@
+use std::convert::TryFrom;
+
+use bytes::BytesMut;
+use flate2::{Compress, Compression, Decompress, FlushCompress, FlushDecompress, Status};
+use headers::WebsocketExtension;
+use http::HeaderValue;
+use thiserror::Error;
+
+use crate::protocol::Role;
+
+const PER_MESSAGE_DEFLATE: &str = "permessage-deflate";
+const CLIENT_NO_CONTEXT_TAKEOVER: &str = "client_no_context_takeover";
+const SERVER_NO_CONTEXT_TAKEOVER: &str = "server_no_context_takeover";
+const CLIENT_MAX_WINDOW_BITS: &str = "client_max_window_bits";
+const SERVER_MAX_WINDOW_BITS: &str = "server_max_window_bits";
+
+const TRAILER: [u8; 4] = [0x00, 0x00, 0xff, 0xff];
+
+/// Errors from `permessage-deflate` extension.
+#[derive(Debug, Error)]
+pub enum DeflateError {
+    /// Compress failed
+    #[error("Failed to compress")]
+    Compress(#[source] std::io::Error),
+    /// Decompress failed
+    #[error("Failed to decompress")]
+    Decompress(#[source] std::io::Error),
+
+    /// Extension negotiation failed.
+    #[error("Extension negotiation failed")]
+    Negotiation(#[source] NegotiationError),
+}
+
+/// Errors from `permessage-deflate` extension negotiation.
+#[derive(Debug, Error)]
+pub enum NegotiationError {
+    /// Unknown parameter in a negotiation response.
+    #[error("Unknown parameter in a negotiation response: {0}")]
+    UnknownParameter(String),
+    /// Duplicate parameter in a negotiation response.
+    #[error("Duplicate parameter in a negotiation response: {0}")]
+    DuplicateParameter(String),
+    /// Received `client_max_window_bits` in a negotiation response for an offer without it.
+    #[error("Received client_max_window_bits in a negotiation response for an offer without it")]
+    UnexpectedClientMaxWindowBits,
+    /// Received unsupported `server_max_window_bits` in a negotiation response.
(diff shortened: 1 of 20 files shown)


daniel-abramov (approved):
I did not check it rigorously, but it looks like a sane implementation, thanks for the amount of work that has been put into it! It also looks like it worked pretty well (based on the comments to the previous iteration of the PR).

It's a significant change, so I'll also additionally ask @agalakhov to approve it so that we can safely merge it!

617a7a:
Any review status on this?

daniel-abramov:
There have not been any comments, so I assume that everyone is okay with the changes, so let's give it a go!

Thanks again for the large amount of work that has been put into that.

nakedible-p:
I was fighting for quite a while trying to get this work and it didn't... until I finally realized that it was merged in *yesterday*! Long time in the making, glad to finally have it implemented.

kazk:
You won't be able to publish `tungstenite` for any other updates/fixes because of the git dependencies. If users don't mind having git dependencies in their apps, they could've used my fork until this was merged.

> It seems like people really want to use permessage-deflate with WebSockets.

Yeah, it can be significant depending on your messages.

> It also looks like it worked pretty well (based on the comments to the previous iteration of the PR).

I've been using the original implementation (#235) in production for almost 2 years without any issues, but the new implementation in this PR is not battle tested. It _should_ be equivalent and have better error handling.

Anyway, I'm just surprised this was merged.

Are you the only active maintainer?

daniel-abramov:
> You won't be able to publish tungstenite for any other updates/fixes because of the git dependencies. If users don't mind having git dependencies in their apps, they could've used my fork until this was merged.

Yeah, that's unfortunate. I may need to move it to a branch (which essentially would defeat the whole purpose of merging I think). It felt like if we don't merge it, it would stagnate and eventually will get obsolete (conflicts with the master branch), but at the same time, I'd forgotten about the git dependencies.

> I've been using the original implementation (https://github.com/snapview/tungstenite-rs/pull/235) in production for almost 2 years without any issues, but the new implementation in this PR is not battle tested. It should be equivalent and have better error handling.

That does not sound too bad.

> Anyway, I'm just surprised this was merged.
> Are you the only active maintainer?

Well, sort of. I'm not the only maintainer, but probably the only one who is more or less active in the past years.

src/extensions/compression/deflate.rs
```diff
@@ -0,0 +1,442 @@
...
+                        ));
+                    }
+
+                    // A server may include the "server_max_window_bits" extension parameter
+                    // in an extension negotiation response even if the extension
+                    // negotiation offer being accepted by the response didn't include the
+                    // "server_max_window_bits" extension parameter.
+                    //
+                    // However, but we need to fail the connection because we don't support it (condition 4).
+                    return Err(DeflateError::Negotiation(
```
nakedible-p, line 258: This is incorrect. Server may include "server_max_window_bits" without the client asking for it for the explicit reason that it informs the window size for compression that the server uses. There is no harm for the client to ignore that value, and use a full sized window for decompression - and since our window is 15 bits by default now that it isn't configurable, any valid value from server is fine for the client.

The correct behaviour is to have the normal duplicate checking for this response header, but accept any valid value (or any value below the window size we use, but flate2 doesn't expose that so we will just need to assume it's 15 bits).

nakedible-p:
@kazk The pull is merged, and the work is stashed in a branch and now there's no open pull for this... but added one review comment that has an impact on the interoperability of the client.

src/extensions/compression/deflate.rs
zy010101: Thanks for your comment.

MrFoxPro:
@kazk https://github.com/hyperium/headers/pull/88#issuecomment-2033009884

bouk:
So it seems someone could rebase this PR on master, but putting the `sec_websocket_extensions.rs` module in this repository directly, since it won't be upstreamed.

SvizelPritula:
I'm working on integrating `sec_websocket_extensions.rs` into this crate, will post a (draft) PR soon.

PR #330, opened 2023-01-12
Switch from base64 0.20 to data-encoding 2

sdroege:
base64 0.21 deprecated the simple API and requires quite a bit more code for the simple usage here, while data-encoding provides exactly the API needed and is also actively maintained.


daniel-abramov:
Nice, thanks!

PR #339, opened 2023-03-19
Make failure adding certificates from rustls_native_certs non-fatal

guusw:
For some reason, loading certain certificates can fail (on my windows machine).
I think kind of failure should be non-fatal and just load the certificates it can

### DIFF
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -70,6 +70,7 @@ mod encryption {
 
     #[cfg(feature = "__rustls-tls")]
     pub mod rustls {
+        use log::warn;
         use rustls::{ClientConfig, ClientConnection, RootCertStore, ServerName, StreamOwned};
 
         use std::{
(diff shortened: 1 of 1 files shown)


daniel-abramov:
Hm. I'm not sure that such behavior would be anticipated by the user.

I.e. I'm afraid that if we silently ignore TLS errors (without returning an error) for a specific TLS backend, this might be quite an unexpected behavior from the user's point of view that may disguise some more serious issues. It feels wrong to just ignore the error inside our library (essentially deciding for the user that "not loaded root certificate is ok"), so I'm leaning towards closing this PR.

I think that if you want to ignore adding root certificates, the code should be explicit about that (i.e. it should not assume that the library will "skip" the errors) - you may want to extract that specific part and run it isolated from `tungstenite` to pinpoint the issue (i.e. to understand why it fails on your side), but if you really have a behavior of ignoring this error, IMHO it's better to implement it in the user code by creating the TLS connection with whatever custom configuration you need and then passing the resulting stream to the [`client()`](https://docs.rs/tungstenite/latest/tungstenite/client/fn.client.html).

PR #350, opened 2023-04-25
Use `rustls-webpki` instead of `webpki`

nickelc:
It hasn't seen a new version in two years.

The `rustls-webpki` crate is a fork of `webpki` which is maintained by the rustls project.


nickelc:
The dependency is not really needed. See #351

PR #358, opened 2023-05-30
Buffer writes before writing to the underlying stream

alexheretic:
Add ability to buffer multiple writes before writing to the underlying stream, controlled by `WebSocketConfig::write_buffer_size` (default 128 KiB). Improves batch message write performance.

This PR comes out of some further investigation this morning regarding writing many small messages. After #357 we no longer flush on each write which gives a big boost. I wondered what would happen if we also avoided/minimised `write` calls to the underlying stream and allowed the messages to buffer in the out_buffer/write buffer.

In short it does seem to provide a boost, though not as dramatic as the flush rework. With #357 my load-generator could write **~2.15M msg/s**, with 128KiB write_buffer_size this goes up to **~2.44M msg/s**.

Larger sizes didn't yield a benefit for me so 128KiB seems a decent default value, I'm proposing that here. The previous code essentially acted as if this value was zero, which could also be a reasonable default though users would need to opt in to get this benefit.

Note: This will only affect users that are calling `write` multiple times then `flush`. Since on `flush` all write buffer data will be written regardless of this config.

### Naming
There is a slight issue with naming clarity here `write_buffer_size` vs `max_write_buffer_size`. I've tried to clarify as much as possible in docs. Perhaps this can be improved?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 - Remove many implicit flushing behaviours. In general reading and writing messages will no 
   longer flush until calling `flush`. An exception is automatic responses (e.g. pongs) 
   which will continue to be written and flushed when reading and writing.
-  This allows writing a batch of messages and flushing once.
+  This allows writing a batch of messages and flushing once, improving performance.
 - Add `WebSocket::read`, `write`, `send`, `flush`. Deprecate `read_message`, `write_message`, `write_pending`.
 - Add `FrameSocket::read`, `write`, `send`, `flush`. Remove `read_frame`, `write_frame`, `write_pending`. 
   Note: Previous use of `write_frame` may be replaced with `send`.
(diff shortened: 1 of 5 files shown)


alexheretic:
> I can probably make a similar micro-bench to show the benefits, or update the current one to include a small simulated io::Write::write cost.

Yep, I re-calibrated the simulated io so underlying `write` calls take **~80ns** and underlying `flush` takes **~8µs**. So still keeping flush as the expensive operation but adding some cost to underlying stream `write`s. This slows `master` code micro-benches to **~18ms** on my setup.

The PR 128 KiB buffer minimises underlying `write` calls and so performs much better.
```
        time:   [6.5226 ms 6.6930 ms 6.7356 ms]
        change: [-63.907% -63.298% -62.689%] (p = 0.07 > 0.05)
```

alexheretic:
@agalakhov wdyt?

agalakhov:
We used to buffer messages and now we're effectively switching to buffering raw data instead. This seems to be a good idea.

My only concern is: what happens if some data get buffered without actually sending them and then a Ping is received? How do we deal with this ordering?

alexheretic:
> My only concern is: what happens if some data get buffered without actually sending them and then a Ping is received? How do we deal with this ordering?

Basically the same as before, since we still auto flush when we schedule a pong response. So on the next read/write/flush if there is any `additional_send` we'll append it to the `out_buffer`. Then flush, that is write all the `out_buffer` to the underlying stream then flush it. The only difference here is the `out_buffer` may have more stuff in it to write, whereas before (or now with `write_buffer_size: 0`) each message would be eagerly written to the underlying stream.

This optimisation is a particularly good fit since we already did write all message frames into the `out_buffer`. The `out_buffer` can already store many messages fine, this is necessary to handle write errors properly. So here all we're doing is delaying underlying stream writes until there are more bytes to write, as this is more efficient.

agalakhov:
Thank you!

PR #359, opened 2023-06-06
remove [] enclosing IPv6 host address

mtbc:
Attempting to use URLs like `ws://[::1]:8080/ws` was giving me "Name or service not known". I adapted https://github.com/sdroege/async-tungstenite/pull/100 to produce this fix which works for me.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -54,6 +54,7 @@ pub fn connect_with_config<Req: IntoClientRequest>(
         let uri = request.uri();
         let mode = uri_mode(uri)?;
         let host = request.uri().host().ok_or(Error::Url(UrlError::NoHostName))?;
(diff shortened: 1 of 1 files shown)


daniel-abramov:
Oh, it seems like something similar has been recently fixed in `tokio-tungstenite`: https://github.com/snapview/tokio-tungstenite/pull/280/files

Did you encounter the problem while using `tungstenite` directly or was it indirectly caused by using `tokio-tungstenite` (that should have had it fixed).

mtbc:
I was using `tungstenite` directly, in https://github.com/btpworks/chronicle/blob/main/crates/gq-subscribe/src/main.rs

daniel-abramov (approved):

PR #368, opened 2023-07-25
Flush the handshake request to ensure that it is sent.

whklhh:
I encountered a problem where the program createing a WebSocket gets stuck as a windows service. The error is that the handshake request was not sent, as observed from the traffic. The HandshakeState machine is trapped in waiting for the response until timeout after 1 minute. 

Then, I checked the dependencies. When sending an HTTP request without flushing using tokio-rustls, it gets stuck.  So I tried adding flushing to the HandshakeState machines, and the problem was solved.

Therefore, I think that adding a flush to the tungstenite handshake may be necessary because tokio-rustls and tokio-tungstenite are both wrappers. However, the root error may be that the flushing process must be done manually instead of automatically in tokio-rustls .

The program has been tested in many environments, but the error only occurs in one of the Windows 11 virtual machine. Additionally, the program also runs correctly when it is not a service in the problematic environment.

### DIFF
--- a/src/handshake/machine.rs
+++ b/src/handshake/machine.rs
@@ -65,18 +65,24 @@ impl<Stream: Read + Write> HandshakeMachine<Stream> {
                 }
             }
             HandshakeState::Writing(mut buf) => {
-                assert!(buf.has_remaining());
-                if let Some(size) = self.stream.write(Buf::chunk(&buf)).no_block()? {
-                    assert!(size > 0);
-                    buf.advance(size);
-                    Ok(if buf.has_remaining() {
-                        RoundResult::Incomplete(HandshakeMachine {
-                            state: HandshakeState::Writing(buf),
-                            ..self
-                        })
+                if self.stream.flush().is_ok() {
+                    if buf.has_remaining() {
+                        if let Some(size) = self.stream.write(Buf::chunk(&buf)).no_block()? {
+                            assert!(size > 0);
+                            buf.advance(size);
+                            Ok(RoundResult::Incomplete(HandshakeMachine {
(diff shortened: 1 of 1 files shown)


daniel-abramov:
Hmm, the change looks somewhat risky and complicated to me. I'm not sure we really need this. I think I would prefer to have an issue created for this (with the description and some reproducible examples so that we can really grasp it).

From your description, it feels like tungstenite is not the right spot to fix the problem (honestly, it feels rather like some sort of bug in the user code, especially if it only happens in some very specific scenario and does not happen otherwise).

CC @agalakhov

whklhh:
I have just compiled the example of tokio-rustls and put the program into the Windows service wrapper. The request cannot be sent until I manually add a flush. And the example of tokio with mio works well. So I think something wrong in tokio-rustls, or it just needs flush manually.
In my opinion, I apologize if I am mistaken, but if the message sent is stuck in the buffer, we should call flush() to forcefully send it.

daniel-abramov:
> In my opinion, I apologize if I am mistaken, but if the message sent is stuck in the buffer, we should call flush to forcefully send it. And flush won't cause any other side effect.

Well, the thing is that it's definitely not stuck in tungstenite 🙂 It may stuck in some other buffer in the underlying stream, but that's something that should be fixed in a different place.

So it seems like there is nothing that indicates that there is a bug in `tungstenite`, and I have not heard any feedback from @agalakhov regarding https://github.com/snapview/tungstenite-rs/pull/368#issuecomment-1649731760, so I hope I can safely assume I can close this PR. If you think this is misguided and you have any new information that would indicate the incorrect behavior of `tungstenite`, feel free to re-open and add new information!

PR #373, opened 2023-08-18
Reduce use of byteorder crate

psychon:
The byteorder dependency is only used in protocol::frame::frame. I thought this dependency could easily be removed and set out to replace the use of byteorder with equivalent std methods.

NetworkEndian is an alias for BigEndian. Converting a number like u32 to bytes can be done via the std lib via .to_be_bytes(). The opposite direction is from_by_bytes(). These simple things thus to not need byteorder.

There is one place in the code where byteorder actually helps, thus this dependency is not actually fully removed. ByteOrder::read_uint() allows to read 1 to 8 bytes of data and returns the result as u64. Doing this with the standard library basically requires re-implementing byteorder. Thus, I did not do that.

-----

Feel free to tell me that this is not useful and reject this. My original motivation was to remove an unnecessary dependency, but apparently that dependency is actually useful, so I failed at my goal.

### DIFF
--- a/src/protocol/frame/frame.rs
+++ b/src/protocol/frame/frame.rs
@@ -1,4 +1,4 @@
-use byteorder::{ByteOrder, NetworkEndian, ReadBytesExt, WriteBytesExt};
+use byteorder::{NetworkEndian, ReadBytesExt};
 use log::*;
 use std::{
     borrow::Cow,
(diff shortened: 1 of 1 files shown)


agalakhov:
Thank you! I believe you're right. At the time of writing the corresponding `std` byteorder operations haven't been in stable Rust and the `byteorder` crate was one and only way to do this without platform-dependent defines.

daniel-abramov (approved):
LGTM! If @agalakhov is fine with the change, we can merge!

agalakhov:
Applied, thanks!

PR #379, opened 2023-09-22
Quick-and-dirty fix for CVE-2023-43669

agalakhov:
This adds header limits in order to reject suspicious requests.

It were much better to use single-pass header parsing instead. Unfortunately, `httparse` does not support it yet and none of its alternatives is mature enough.


src/error.rs
```diff
@@ -59,6 +59,9 @@ pub enum Error {
     /// UTF coding error.
     #[error("UTF-8 encoding error")]
     Utf8,
+    /// Attack attempt detected.
+    #[error("Attack attempt detected")]
+    AttackAttempt,
```
daniel-abramov, line 64: I think we could use a bit more descriptive name, perhaps something like `HeadersTooLarge(usize)` or similar, otherwise, it may be a bit hard to tell what went wrong in a rare case when the user bug (an invalid client that sends large headers for some obscure reasons).

src/handshake/machine.rs
```diff
@@ -20,7 +20,10 @@ pub struct HandshakeMachine<Stream> {
 impl<Stream> HandshakeMachine<Stream> {
     /// Start reading data from the peer.
     pub fn start_read(stream: Stream) -> Self {
-        HandshakeMachine { stream, state: HandshakeState::Reading(ReadBuffer::new()) }
+        HandshakeMachine {
```
daniel-abramov, line 23:
```suggestion
        Self {
```

src/handshake/machine.rs
```diff
@@ -119,7 +128,54 @@ pub trait TryParse: Sized {
...
-    Reading(ReadBuffer),
+    Reading(ReadBuffer, AttackCheck),
     /// Sending data to the peer.
     Writing(Cursor<Vec<u8>>),
 }
+
+/// Attack mitigation. Contains counters needed to prevent DoS attacks
+/// and reject valid but useless headers.
+#[derive(Debug)]
+pub(crate) struct AttackCheck {
```
daniel-abramov, line 139: A useful comment, but I think we can't make an assumption that these were "valid but useless headers". If I got the implementation right, it guards against sending too many packets (which contain headers in our case) or too large header(s). Maybe the name could be something like `HeaderSizeGuard` (I don't have any strong preference though).

src/handshake/machine.rs
```diff
@@ -119,7 +128,54 @@ pub trait TryParse: Sized {
...
+    /// Number of HTTP header successful reads (TCP packets).
+    number_of_packets: usize,
+    /// Total number of bytes in HTTP header.
+    number_of_bytes: usize,
+}
+
+impl AttackCheck {
+    /// Initialize attack checking for incoming buffer.
+    fn new() -> Self {
+        AttackCheck { number_of_packets: 0, number_of_bytes: 0 }
```
daniel-abramov, line 149:
```suggestion
        Self { number_of_packets: 0, number_of_bytes: 0 }
```

src/handshake/machine.rs
```diff
@@ -119,7 +128,54 @@ pub trait TryParse: Sized {
...
+    }
+
+    /// Check size of incoming packet. To be called immediately after `read()`
+    /// passing its return bytes count as `size`.
```
daniel-abramov, line 153:
```suggestion
    /// Check the size of an incoming packet. To be called immediately after `read()`
    /// passing its returned bytes count as `size`.
```

src/handshake/machine.rs
```diff
@@ -119,7 +128,54 @@ pub trait TryParse: Sized {
...
+
+        if self.number_of_packets > MAX_PACKETS {
+            return Err(Error::AttackAttempt);
+        }
+
+        if self.number_of_packets > MIN_PACKET_CHECK_THRESHOLD {
+            if self.number_of_packets * MIN_PACKET_SIZE > self.number_of_bytes {
+                return Err(Error::AttackAttempt);
+            }
+        }
```
daniel-abramov, line 177: I understand the meaning of the first 2 checks, I'm not sure what's the exact reason behind the last one though.

If I get it right, the main purpose of the component is to fail once "too many" reads from a socket occur or in case the total size of the header section gets too large.

daniel-abramov (approved):
Thanks for tackling this issue! I've left a couple of minor comments, but overall it looks good for a quick fix!

src/error.rs
agalakhov: This is intended to use for any suspicious activity, not only for headers that are too large. For example, sending a medium-size header in 1-byte packets is bad for the current `httparse` and results in the same error. We could deprecate and rename this later after improving HTTP parsing in general.

src/handshake/machine.rs
agalakhov: Header that are too large aren't the only "useless" ones. We use `httparse` for parsing, and `httparse` is not able to resume incomplete header parsing. Since we "do not know anything" about headers format (i.e. don't want to search for CRLFCRLF on our own), we run `httparse` multiple times. This results in O(n^2) in worst case and that's exactly was resulted in CVE-2023-43669. There is one more way to trigger the same behavior: just send usual headers in 1-byte TCP packets in such a way that `read()` always returns one byte. This is taken into account as well.

src/handshake/machine.rs
agalakhov: This is to prevent O(n^2) parsing in case somebody sends us headers of reasonable size in very small (i.e. 1-byte) chunks.

PR #391, opened 2023-12-01
Fix `FrameHeader::format` write & up minimum rustls to `0.21.6`

alexheretic:
* Fix `FrameHeader::format` write correctness, should use `write_all` (though the small writes would be unlikely to be an issue). Thanks clippy!
* Up minimum _rustls_ to `0.21.6` since `RootCertStore::add_trust_anchors` is used here and was introduced in that version.
* Also addressed a few other more minor clippy lints.


daniel-abramov (approved):
Awesome. Thanks for updating the changelog as well!

alexheretic:
> Awesome. Thanks for updating the changelog as well!

I put "(0.20.2)" in there as these are non-breaking changes, but I just noticed this is wrong, as the _http_ crate bump is breaking so presumably the next version will be a breaking bump regardless.

PR #393, opened 2023-12-02
Fix auto pong responses not flushing after block

alexheretic:
Currently auto pongs are written and flushed during `read()`. If a `WouldBlock` error is received it is ignored and reading continues.

In the case the usage is read-predominant, this presents a scenario where the auto pong is `WouldBlock`-ed and will never retry flushing until another ping is received or the client does some unrelated writes. This could cause the other connected party to time out the connection as it has failed to receive a timely pong response.

I reproduced this in new test `read_usage_auto_pong_flush`.

To fix we record if a pong write failed in this way and retry on all subsequent `read` calls until a flush succeeds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Unreleased
+- Fix read-predominant auto pong responses not flushing when hitting WouldBlock errors.
 - Improve `FrameHeader::format` write correctness.
 - Up minimum _rustls_ to `0.21.6`.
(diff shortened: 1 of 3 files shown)


daniel-abramov (approved):
Good catch! Thanks!

NB I've now [again] realized that we need to rewrite our state machine (the code is getting increasingly convoluted and hard to manage) (edit: not the fault of this PR, just a general remark).

PR #397, opened 2023-12-08
Update MSRV to 1.60 and check on the CI

sdroege:
The code does not actually compile anymore with 1.51.

Also only run `cargo check` with 1.60 as various optional features, tests, benchmarks actually require newer Rust versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,30 @@ jobs:
       - name: Test
         run: cargo test --release
 
+  test-msvr:
+    name: Test MSRV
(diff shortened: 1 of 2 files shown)


sdroege:
Doesn't look like `cargo hack` even works with that version? :)

sdroege:
Also 1.60 only works with default features and not any of the tests. So I'll limit the CI to that I guess.

sdroege:
Only running `cargo check` now with 1.60 as other (optional) things actually require newer Rust versions than that.

sdroege:
@daniel-abramov All passing it seems.

PR #400, opened 2024-01-10
Add builder for additional header values

ManevilleF:
## Work done

Added a `ClientRequestBuilder` type inspired by [websockets](https://docs.rs/websockets/latest/websockets/struct.WebSocketBuilder.html) to allow for additional header values in the request, like authorization or subprotocols.

The way I implemented this is not the cleanest but I wanted to avoid any breaking change.

This is merely a proposal, so feedback welcome !

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -274,6 +274,8 @@ impl<'h, 'b> IntoClientRequest for httparse::Request<'h, 'b> {
 ///
 /// ```rust no_run
 /// # use crate::*;
+/// use http::Uri;
+/// use tungstenite::{connect, ClientRequestBuilder};
 ///
 /// let uri: Uri = "ws://localhost:3012/socket".parse().unwrap();
 /// let token = "my_jwt_token";
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,12 +1,13 @@
 //! Methods to connect to a WebSocket as a client.
 
 use std::{
+    convert::TryFrom,
     io::{Read, Write},
     net::{SocketAddr, TcpStream, ToSocketAddrs},
     result::Result as StdResult,
 };
 
-use http::{request::Parts, Uri};
+use http::{request::Parts, HeaderName, Uri};
 use log::*;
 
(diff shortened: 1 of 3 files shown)


Its-Just-Nans:
This would be great to simulate a connection from a browser

- <https://datatracker.ietf.org/doc/html/rfc6455>

```text
8.   The request MUST include a header field with the name |Origin|
        [[RFC6454](https://datatracker.ietf.org/doc/html/rfc6454)] if the request is coming from a browser client.  If
        the connection is from a non-browser client, the request MAY
        include this header field if the semantics of that client match
        the use-case described here for browser clients.  The value of
        this header field is the ASCII serialization of origin of the
        context in which the code establishing the connection is
        running.  See [[RFC6454](https://datatracker.ietf.org/doc/html/rfc6454)] for the details of how this header field
        value is constructed.
```

daniel-abramov (changes requested):
NB: These changes might indeed be useful, but the CI pipeline fails. I also think writing a couple of tests for the subprotocol functionality would be useful.

Its-Just-Nans:
@ManevilleF  I'v made PR once again on your repo to correct doc (once again 😅) and add a test

Link to the PR: https://github.com/ManevilleF/tungstenite-rs/pull/2

Maybe after you merged it @daniel-abramov could merge this one!

daniel-abramov:
Nice. Btw, one of the CI steps fails. I did not check it in a detail, but my first guess is that the new example [might require enabling some features](https://github.com/snapview/tungstenite-rs/blob/master/Cargo.toml#L78).

Its-Just-Nans:
@ManevilleF  If you add
```
#![cfg(feature = "handshake")]
```
at the top of [the test file](https://github.com/ManevilleF/tungstenite-rs/blob/4e88b43f93aed436f2e1057aaf49a676c4016822/tests/client_headers.rs#L1)

I guess it should work ?

daniel-abramov (approved):

Issue #412, opened 2024-03-06
Remove or make even more optional the `url` crate in the next version. 


Thomas-airmatrix:
Outside of tests there is only a single use of `url::Url` in tungstenite:
```rust
impl<'a> IntoClientRequest for &'a Url {
    fn into_client_request(self) -> Result<Request> {
        self.as_str().into_client_request()
    }
}

impl IntoClientRequest for Url {
    fn into_client_request(self) -> Result<Request> {
        self.as_str().into_client_request()
    }
}
```
Which is just implementing some helper conversion traits.

Instead, we can just have users call `as_str()` on the Url directly if needed.

With `http` =1,  `http::uri::Uri` has replaced most uses of the `url` crate. I have multiple project such that the only use of `url` is tungstenite.

Its-Just-Nans:
This can be closed ?!

Issue #409, opened 2024-03-01
Bump version


Its-Just-Nans:
Hello,

Is it possible to create a PR to bump the version to release the header upgrade ?

Or should I wait for https://github.com/snapview/tungstenite-rs/pull/363 to be merged (then the version will be hopefully upgraded) ?

daniel-abramov:
Hm, I don't have a strong opinion on that, but it looks like the change introduced in https://github.com/snapview/tungstenite-rs/pull/400 is the only one since the latest release. Let's see if there is any progress on #363 if that's fine.

PR #363, opened 2023-06-28
Subprotocol header

conorbros:
Add client-side handling for subprotocol header during handshake.

PR #419, opened 2024-04-21
Make url as feature and bump version

Its-Just-Nans:
Hello,

This PR fixes:

- https://github.com/snapview/tungstenite-rs/issues/412
- https://github.com/snapview/tungstenite-rs/issues/409


daniel-abramov (commented):
Looks good. One question though: why is the newly introduced `tests/url_feature.rs` is a copy-paste of the existing test (with the only single exception of using `Url`)? - I think this might be a bit confusing.

Its-Just-Nans:
Yes.
I changed the test. Is that better ?

Thanks

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.22.0
+- Make `url::Url` as feature
+- Add Header and subprotocol support at creation
```
daniel-abramov, line 3:
```suggestion
- Make `url` optional.
- Add a builder for convenient headers and subprotocols construction.
```

Cargo.toml
```diff
@@ -9,7 +9,7 @@ readme = "README.md"
 homepage = "https://github.com/snapview/tungstenite-rs"
 documentation = "https://docs.rs/tungstenite/0.21.0"
```
daniel-abramov, line 10:
```suggestion
documentation = "https://docs.rs/tungstenite/0.22.0"
```

If the version is raised, we must also update the documentation link.

daniel-abramov (changes requested):
Looks good!

Just to clarify one thing: if I got your intentions right, you want us to publish a new version after merging this PR, right? (otherwise CHANGELOG should contain `Unreleased` instead of `0.22.0` and also the update in the `Cargo.toml` would also not be necessary).

AFAIR originally we wanted to first wait for https://github.com/snapview/tungstenite-rs/pull/363 first and then release a version. But if a builder pattern is useful on its own, we could make a release before #363 I think.

Its-Just-Nans:
> Looks good!
> 
> Just to clarify one thing: if I got your intentions right, you want us to publish a new version after merging this PR, right? (otherwise CHANGELOG should contain `Unreleased` instead of `0.22.0` and also the update in the `Cargo.toml` would also not be necessary).
> 
> AFAIR originally we wanted to first wait for #363 first and then release a version. But if a builder pattern is useful on its own, we could make a release before #363 I think.

Yes, make a release version is my goal. With a new version I could (update and) merge https://github.com/rerun-io/ewebsock/pull/27 crate which is using tungstenite

For #363 I think @conorbros is a little busy at the moment because I sent him a PR (https://github.com/shotover/tungstenite-rs/pull/1) but he didn't accept it for now

A solution could be to merge https://github.com/Its-Just-Nans/tungstenite-rs/tree/master directly into tungstenite (without passing by the fork of @conorbros). But that would make @conorbros invalid his PR. I don't know if it's a good and acceptable way of doing it

Its-Just-Nans:
I think #363 may be mergeable now (thank you @conorbros) !

So maybe we can merge #363 then this one ?

daniel-abramov:
That makes sense. I think we could merge #363; it just requires minor updates to pass the CI.

PR #425, opened 2024-05-13
Update rustls

rukai:
Just wanted to make sure the rustls update got in before tungstenite 0.22 was released.


daniel-abramov (commented):
Looks like it needs a couple more changes to compile.

rukai:
Ah unfortunate, env_logger MSRV is to high for this project, I've reverted the change for env_logger.
rustls is the only one that really matters anyway.

PR #426, opened 2024-05-16
Add `permessage-deflate` support, again

SvizelPritula:
In #328, @kazk implemented support for the `permessage-deflate` extention. Implementing this extention required the ability to parse the `Sec-Websocket-Extentions` header, so @kazk submitted a [pull request to the `headers` crate](https://github.com/hyperium/headers/pull/88). However, the maintainers of the `headers` crate don't wan't to support this header yet, as they don't wan't to commit to any particular API. As such, [they suggested](https://github.com/hyperium/headers/pull/88#issuecomment-2033009884) that it should be implemented in this or some other crate, tweaked as necessary and potentially moved into the `headers` crate in the future.

This PR reverts the commit that reverted @kazk's commit that added `permessage-deflate` support. It also copies his implementation of `SecWebsocketExtentions` intended for the `headers` crate.

The implementation of `SecWebsocketExtentions` relied on some internal utilities of the `headers` crate. I've removed some of those dependencies and re-implemented others. I have also gated the implementation behind the existing `handshake` feature, as well as any code that relies on it, since it relies on the `headers` crate, which in turn relies on many other HTTP crates.

## Blocking issues

- @nakedible-p found [an error](https://github.com/snapview/tungstenite-rs/pull/328#discussion_r1167592235) in the implementation. As I know very little about the Websocket spec, I'm unsure if I can fix it properly.
- The `headers` crate is licensed under the MIT license, while this crate uses a dual MIT+Apache license. Assuming we don't want to change the license on this crate, @kazk would need to consent to the re-licensing of his PR.

## Unresolved questions

- It is impossible to use the `deflate` feature without having `tungstenite` handle the Websocket handshake. There is a `WebSocket::from_raw_socket_with_extensions` method to allow for exactly that, but it takes an `Extensions` struct, which has no public constructor.
- The only way to get such a struct is from the `WebSocketConfig::accept_offers` method, which implements the server part of extension negotiation and requires the `handshake` feature. This method is intended to allow for `tungstenite` to be integrated into web frameworks. No similar public method exists for clients, although that might not be a big issue, since a client knows whether a request will be a Websocket request up front.
- To me, the `WebSocketConfig::accept_offers` feels out of place on the `WebSocketConfig`. We could make it a standalone function, perhaps in the `extensions` module.
- I have made the `SecWebsocketExtentions` implementation public. This is necessary to make the `WebSocketConfig::accept_offers` method usable. This does mean that if `SecWebsocketExtentions` were to be added to `headers` in the future, switching to that implementation would (I think) be a breaking change. An alternative would be to have `WebSocketConfig::accept_offers` take and return a `HeaderValue`, which it would parse itself. This would allow us to make `SecWebsocketExtentions` a private implementation detail.
- This is a breaking change, since a new field was added to `WebSocketConfig` and a new variant was added to the `Error` enum. Since this field and variant depends on the `deflate` feature, I've annotated both with `#[non_exhaustive]`. I've done the same to `ProtocolError`, which currently has variants gated behind `handshake`. It would be possible to disable `#[non_exhaustive]` if all relevant features are enabled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 target
 Cargo.lock
 .vscode
+autobahn/client/
+autobahn/server/
--- a/src/extensions/compression/deflate.rs
+++ b/src/extensions/compression/deflate.rs
@@ -1,0 +1,442 @@
+use std::convert::TryFrom;
+
+use bytes::BytesMut;
+use flate2::{Compress, Compression, Decompress, FlushCompress, FlushDecompress, Status};
+use headers::WebsocketExtension;
+use http::HeaderValue;
+use thiserror::Error;
+
+use crate::protocol::Role;
+
+const PER_MESSAGE_DEFLATE: &str = "permessage-deflate";
+const CLIENT_NO_CONTEXT_TAKEOVER: &str = "client_no_context_takeover";
+const SERVER_NO_CONTEXT_TAKEOVER: &str = "server_no_context_takeover";
+const CLIENT_MAX_WINDOW_BITS: &str = "client_max_window_bits";
+const SERVER_MAX_WINDOW_BITS: &str = "server_max_window_bits";
+
+const TRAILER: [u8; 4] = [0x00, 0x00, 0xff, 0xff];
+
+/// Errors from `permessage-deflate` extension.
+#[derive(Debug, Error)]
+pub enum DeflateError {
+    /// Compress failed
+    #[error("Failed to compress")]
+    Compress(#[source] std::io::Error),
+    /// Decompress failed
+    #[error("Failed to decompress")]
+    Decompress(#[source] std::io::Error),
+
+    /// Extension negotiation failed.
+    #[error("Extension negotiation failed")]
+    Negotiation(#[source] NegotiationError),
+}
+
+/// Errors from `permessage-deflate` extension negotiation.
+#[derive(Debug, Error)]
+pub enum NegotiationError {
+    /// Unknown parameter in a negotiation response.
+    #[error("Unknown parameter in a negotiation response: {0}")]
+    UnknownParameter(String),
+    /// Duplicate parameter in a negotiation response.
+    #[error("Duplicate parameter in a negotiation response: {0}")]
+    DuplicateParameter(String),
+    /// Received `client_max_window_bits` in a negotiation response for an offer without it.
+    #[error("Received client_max_window_bits in a negotiation response for an offer without it")]
+    UnexpectedClientMaxWindowBits,
+    /// Received unsupported `server_max_window_bits` in a negotiation response.
+    #[error("Received unsupported server_max_window_bits in a negotiation response")]
+    ServerMaxWindowBitsNotSupported,
+    /// Invalid `client_max_window_bits` value in a negotiation response.
+    #[error("Invalid client_max_window_bits value in a negotiation response: {0}")]
+    InvalidClientMaxWindowBitsValue(String),
+    /// Invalid `server_max_window_bits` value in a negotiation response.
+    #[error("Invalid server_max_window_bits value in a negotiation response: {0}")]
+    InvalidServerMaxWindowBitsValue(String),
+    /// Missing `server_max_window_bits` value in a negotiation response.
+    #[error("Missing server_max_window_bits value in a negotiation response")]
+    MissingServerMaxWindowBitsValue,
+}
+
+// Parameters `server_max_window_bits` and `client_max_window_bits` are not supported for now
+// because custom window size requires `flate2/zlib` feature.
+/// Configurations for `permessage-deflate` Per-Message Compression Extension.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct DeflateConfig {
+    /// Compression level.
+    pub compression: Compression,
+    /// Request the peer server not to use context takeover.
+    pub server_no_context_takeover: bool,
+    /// Hint that context takeover is not used.
+    pub client_no_context_takeover: bool,
+}
+
+impl DeflateConfig {
+    pub(crate) fn name(&self) -> &str {
+        PER_MESSAGE_DEFLATE
+    }
+
+    /// Value for `Sec-WebSocket-Extensions` request header.
+    pub(crate) fn generate_offer(&self) -> WebsocketExtension {
+        let mut offers = Vec::new();
+        if self.server_no_context_takeover {
+            offers.push(HeaderValue::from_static(SERVER_NO_CONTEXT_TAKEOVER));
+        }
+
+        // > a client informs the peer server of a hint that even if the server doesn't include the
+        // > "client_no_context_takeover" extension parameter in the corresponding
+        // > extension negotiation response to the offer, the client is not going
+        // > to use context takeover.
+        // > https://www.rfc-editor.org/rfc/rfc7692#section-7.1.1.2
+        if self.client_no_context_takeover {
+            offers.push(HeaderValue::from_static(CLIENT_NO_CONTEXT_TAKEOVER));
+        }
+        to_header_value(&offers)
+    }
+
+    /// Returns negotiation response based on offers and `DeflateContext` to manage per message compression.
+    pub(crate) fn accept_offer(
+        &self,
+        offers: &headers::SecWebsocketExtensions,
+    ) -> Option<(WebsocketExtension, DeflateContext)> {
+        // Accept the first valid offer for `permessage-deflate`.
+        // A server MUST decline an extension negotiation offer for this
+        // extension if any of the following conditions are met:
+        // 1. The negotiation offer contains an extension parameter not defined for use in an offer.
+        // 2. The negotiation offer contains an extension parameter with an invalid value.
+        // 3. The negotiation offer contains multiple extension parameters with the same name.
+        // 4. The server doesn't support the offered configuration.
+        offers.iter().find_map(|extension| {
+            if let Some(params) = (extension.name() == self.name()).then(|| extension.params()) {
+                let mut config =
+                    DeflateConfig { compression: self.compression, ..DeflateConfig::default() };
+                let mut agreed = Vec::new();
+                let mut seen_server_no_context_takeover = false;
+                let mut seen_client_no_context_takeover = false;
+                let mut seen_client_max_window_bits = false;
+                for (key, val) in params {
+                    match key {
+                        SERVER_NO_CONTEXT_TAKEOVER => {
+                            // Invalid offer with multiple params with same name is declined.
+                            if seen_server_no_context_takeover {
+                                return None;
+                            }
+                            seen_server_no_context_takeover = true;
+                            config.server_no_context_takeover = true;
+                            agreed.push(HeaderValue::from_static(SERVER_NO_CONTEXT_TAKEOVER));
+                        }
+
+                        CLIENT_NO_CONTEXT_TAKEOVER => {
+                            // Invalid offer with multiple params with same name is declined.
+                            if seen_client_no_context_takeover {
+                                return None;
+                            }
+                            seen_client_no_context_takeover = true;
+                            config.client_no_context_takeover = true;
+                            agreed.push(HeaderValue::from_static(CLIENT_NO_CONTEXT_TAKEOVER));
+                        }
+
+                        // Max window bits are not supported at the moment.
+                        SERVER_MAX_WINDOW_BITS => {
+                            // Decline offer with invalid parameter value.
+                            // `server_max_window_bits` requires a value in range [8, 15].
+                            if let Some(bits) = val {
+                                if !is_valid_max_window_bits(bits) {
+                                    return None;
+                                }
+                            } else {
+                                return None;
+                            }
+
+                            // A server declines an extension negotiation offer with this parameter
+                            // if the server doesn't support it.
+                            return None;
+                        }
+
+                        // Not supported, but server may ignore and accept the offer.
+                        CLIENT_MAX_WINDOW_BITS => {
+                            // Decline offer with invalid parameter value.
+                            // `client_max_window_bits` requires a value in range [8, 15] or no value.
+                            if let Some(bits) = val {
+                                if !is_valid_max_window_bits(bits) {
+                                    return None;
+                                }
+                            }
+
+                            // Invalid offer with multiple params with same name is declined.
+                            if seen_client_max_window_bits {
+                                return None;
+                            }
+                            seen_client_max_window_bits = true;
+                        }
+
+                        // Offer with unknown parameter MUST be declined.
+                        _ => {
+                            return None;
+                        }
+                    }
+                }
+
+                Some((to_header_value(&agreed), DeflateContext::new(Role::Server, config)))
+            } else {
+                None
+            }
+        })
+    }
+
+    pub(crate) fn accept_response<'a>(
+        &'a self,
+        agreed: impl Iterator<Item = (&'a str, Option<&'a str>)>,
+    ) -> Result<DeflateContext, DeflateError> {
+        let mut config = DeflateConfig {
+            compression: self.compression,
+            // If this was hinted in the offer, the client won't use context takeover
+            // even if the response doesn't include it.
+            // See `generate_offer`.
+            client_no_context_takeover: self.client_no_context_takeover,
+            ..DeflateConfig::default()
+        };
+        let mut seen_server_no_context_takeover = false;
+        let mut seen_client_no_context_takeover = false;
+        // A client MUST _Fail the WebSocket Connection_ if the peer server
+        // accepted an extension negotiation offer for this extension with an
+        // extension negotiation response meeting any of the following
+        // conditions:
+        // 1. The negotiation response contains an extension parameter not defined for use in a response.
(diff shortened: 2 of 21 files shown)


nakedible-p:
I will be happy to go through the details of the permessage-deflate spec – or rather even fix the bug myself, while some missing window bits support. However, the bug is extremely niche as these extension fields are very rarely used, so I'd prefer to try to get this merged and then do an improvement pull on top of that.

SvizelPritula:
@kazk Would you be willing to relicense your [PR](https://github.com/hyperium/headers/pull/88) under the MIT + Apache dual license as used by this crate?

If not, I believe we can use the code under MIT + Apache anyways, as long as we add a `NOTICE` file to the repo that looks a little bit like this:

```
This crate contains code copied from the headers crate, licensed as follows:

*Insert copy of headers license here.*
```

Additionaly, we would probably have to add "Copyright (c) 2014-2023 Sean McArthur" to `LICENSE-MIT`, which is odd, given that no part of the code in this PR was actually written by Sean McArthur.

zitsen:
What's the status of this pr?

SvizelPritula:
I'm not a lawyer, but I *think* that the notices I've to the license files should be compliant with all relevant licenses. As such, this PR is now ready for review! :partying_face: Sorry for the delay.

kazk:
@SvizelPritula 

I apologize for the delay. 

> @kazk Would you be willing to relicense your https://github.com/hyperium/headers/pull/88 under the MIT + Apache dual license as used by this crate?

Yes, you can do whatever is necessary for what I did :)
Thanks for your work!

zitsen:
Glad to see it still going on, thanks your great work @SvizelPritula @kazk .

So are there any other problems pending this?

Sorry for disturbing, but seems that I should ping you @nakedible-p @daniel-abramov .

daniel-abramov:
I remember I checked https://github.com/snapview/tungstenite-rs/pull/328 back then, and it seems like there are many common parts ([back then it was reverted due to dependencies](https://github.com/snapview/tungstenite-rs/pull/328#issuecomment-1480370143). So yeah, I'm not against merging it.

I'd also appreciate it if someone who tracked the issue could approve it, though (to ensure that I did not miss anything since I did not go as thoroughly through all the changes as I typically try to do).

daniel-abramov (approved):

goriunov:
We have been running a copy of this branch for couple of weeks now, at least in the standard cases it seems to work good, both directions server client and client server different languages

daniel-abramov:
> We have been running a copy of this branch for couple of weeks now in production, at least in the standard cases it seems to work good, both directions server client and client server, different languages clients connect to the server.

This sounds pretty good! It's always good to know that someone tested it in production!

@SvizelPritula, would you be interested in rebasing on top of a `master` branch? (there have been quite significant changes recently that affect performance and change the API surface a bit)

goriunov:
@daniel-abramov @SvizelPritula just checking in. Is there anything else that needs to be done before this branch can be merged?

SvizelPritula:
> just checking in. Is there anything else that needs to be done before this branch can be merged?

I've merged this branch with `master`, which broke the Autobahn test suite. I'll try to investigate why.

goriunov:
Just checking in if there are any updates on this PR?

SvizelPritula:
The tests are now fixed, everything *should* be working.

Sorry for taking so long, I was busy with exams.

daniel-abramov (commented):
Thanks for the update!

I had to quickly go through the changes again (albeit skipping all deflate-specific logic) as I keep forgetting the things between reviews (the PR is pretty large).

Generally it looks good, I'm only a bit concerned with a slightly more complicated logic in `protocol/**`  (we have more branching and more conditions), but making it better would possibly require a bigger overhaul.

I also decided to run our benchmarks locally and it seems like this version is somewhat slower than the version in `master` when it comes to reading. I consistently get regression of around 15-17%.

```
                        time:   [6.6915 ms 6.7087 ms 6.7273 ms]
                        change: [+14.968% +15.822% +16.508%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 17 outliers among 100 measurements (17.00%)
  6 (6.00%) low mild
  5 (5.00%) high mild
  6 (6.00%) high severe

                        time:   [6.4633 ms 6.4821 ms 6.5024 ms]
                        change: [+17.048% +17.722% +18.370%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
```

Unfortunately, I did not have time to dig in and investigate the root causes, but by quickly going through `protocol/**` changes I assume that it might be related to the changes around `IncompleteMessage` / `extend_incomplete()`.

.gitignore
```diff
@@ -1,3 +1,5 @@
 target
 Cargo.lock
 .vscode
+autobahn/client/
+autobahn/server/
```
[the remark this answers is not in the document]
SvizelPritula: Those folders contain the test results after running `scripts/autobahn-client.sh` or `scripts/autobahn-server.sh`. Those shouldn't be commited, so I added them to `.gitignore`.

zitsen:
Any update? @SvizelPritula  @daniel-abramov

SvizelPritula:
> Any update? @SvizelPritula  @daniel-abramov 

I have tried to investigate the regression a bit, but I haven't been successful yet.

This PR is very old, and the crate has seen a lot of performance improvements since it began. (In fact, IIRC this PR is older than the benchmark that regressed.) This unfortunately makes it difficult to find out which change exactly is to blame.

goriunov:
Hey, wondering if there is any progress on this PR? 

Would be great to get it in main at some point. We are currently running some production builds from older version of this branch, so not very ideal.

unfortunately we can not run system without compression as we have very very traffic dependent systems that streams terabytes per months and if we remove compression AWS bill goes crazy.

unfortunately there is no proper tested alternative in rust websockets that has compression built in, so we kinda have to hack around

daniel-abramov:
@goriunov, I can understand that. However, merging it "as is" is a bit complicated because there is a minor performance degradation, and no one has had time to investigate its cause. I would be somewhat uncomfortable merging it in the current state, knowing that it may cause some issues that I won't be able to debug/solve quickly (it's a pretty large PR after all).

I thought about it for a bit, and I think one of the ways to approach the problem would be as follows:
1. 70% of the code is deflate and WebSocket extension parsing. Perhaps this could be merged separately under a feature gate or be moved to a separate crate (unless there is already one for it), as it may also help other websocket crates that might want to implement the feature.
2. Then, the rest of the code could be submitted as separate (smaller) PR (or PRs!) that are easier to review and discuss. This would also allow us to see at which point the performance degradation starts to happen and debug it appropriately.

Unfortunately I did not participate that much in the development of this feature and could not take it over as I did not really use deflate in my projects and I rarely use WebSockets nowadays, so I always struggled to find time to justify the effort (I believe it would probably take a couple of days to properly go through changes one more time, merge them one-by-one, investigate the issues, benchmark, etc).

SvizelPritula:
I think I managed to resolve the performance regression.

Unfortunately, the performance of the benchmarks still regresses when the `deflate` feature is enabled - but that at least doesn't impact existing users.

src/extensions/compression/deflate.rs
```diff
@@ -0,0 +1,456 @@
...
+
+#[cfg(feature = "handshake")]
+impl DeflateConfig {
+    pub(crate) fn name(&self) -> &str {
+        PER_MESSAGE_DEFLATE
+    }
+
+    /// Value for `Sec-WebSocket-Extensions` request header.
+    pub(crate) fn generate_offer(&self) -> WebsocketExtension {
+        let mut offers = Vec::new();
```
erebe, line 90: The allocation can be avoided by unrolling the if

```rust
        // > a client informs the peer server of a hint that even if the server doesn't include the
        // > "client_no_context_takeover" extension parameter in the corresponding
        // > extension negotiation response to the offer, the client is not going
        // > to use context takeover.
        // > https://www.rfc-editor.org/rfc/rfc7692#section-7.1.1.2
        match (self.server_no_context_takeover, self.client_no_context_takeover) {
            (true, true) => to_header_value(&[HeaderValue::from_static(SERVER_NO_CONTEXT_TAKEOVER), HeaderValue::from_static(CLIENT_NO_CONTEXT_TAKEOVER)]),
            (true, false) => to_header_value(&[HeaderValue::from_static(SERVER_NO_CONTEXT_TAKEOVER)]),
            (false, true) => to_header_value(&[HeaderValue::from_static(CLIENT_NO_CONTEXT_TAKEOVER)]),
            (false, false) => to_header_value(&[]), 
        }
```

src/extensions/compression/deflate.rs
```diff
@@ -0,0 +1,456 @@
...
+        if is_final && !self.peer_context_takeover() {
+            self.decompressor.reset(false);
+        }
+
+        Ok(output)
+    }
+}
+
+#[cfg(feature = "handshake")]
+fn to_header_value(params: &[HeaderValue]) -> WebsocketExtension {
```
erebe, line 447: It can be speeded up for the nominal case, where params is empty.
Instead of creating each time a new buffer and trying to parse the header for holding the known value `permessage-deflate`

```rust

if params.is_empty() {
   return WebsocketExtension::default()
}

// Pay the cost of creating a new buffer, creating header value, and parsing it for websocket extension
```

src/extensions/compression/deflate.rs
```diff
@@ -0,0 +1,456 @@
...
+            Role::Client => !self.config.server_no_context_takeover,
+        }
+    }
+
+    // Compress the data of message.
+    pub(crate) fn compress(&mut self, data: &[u8]) -> Result<Vec<u8>, DeflateError> {
+        // https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.1
+        // 1. Compress all the octets of the payload of the message using DEFLATE.
+        let mut output = Vec::with_capacity(data.len());
+        let before_in = self.compressor.total_in() as usize;
```
erebe, line 375: You should not cast u64 into a usize, as on 32bits arch this can overflow.
Instead turn the `usize` of data.len() into an u64 which is safer (as 128bits arch is not really a thing)

erebe, line 375: You should not cast u64 into a usize, as on 32bits arch this can overflow.
Instead turn the `usize` of data.len() into an u64 which is safer (as 128bits arch is not really a thing)

the offset can be converted into usize more safely, or use `usize::try_from()`

src/extensions/compression/deflate.rs
```diff
@@ -0,0 +1,456 @@
...
+            {
+                Status::Ok | Status::BufError => continue,
+                Status::StreamEnd => break,
+            }
+        }
+        // 3. Remove 4 octets (that are 0x00 0x00 0xff 0xff) from the tail end.
+        //    After this step, the last octet of the compressed data contains
+        //    (possibly part of) the DEFLATE header bits with the "BTYPE" bits
+        //    set to 00.
+        output.truncate(output.len() - 4);
```
erebe, line 406: `4` can be replaced by `TRAILER.lend()`

src/extensions/compression/deflate.rs
```diff
@@ -0,0 +1,456 @@
...
+    pub(crate) fn decompress(
+        &mut self,
+        mut data: Vec<u8>,
+        is_final: bool,
+    ) -> Result<Vec<u8>, DeflateError> {
+        if is_final {
+            data.extend_from_slice(&TRAILER);
+        }
+
+        let before_in = self.decompressor.total_in() as usize;
```
erebe, line 424: same comment regarding turning u64 into a usize

src/extensions/compression/deflate.rs
```diff
@@ -0,0 +1,456 @@
...
+        Ok(output)
+    }
+
+    pub(crate) fn decompress(
+        &mut self,
+        mut data: Vec<u8>,
+        is_final: bool,
+    ) -> Result<Vec<u8>, DeflateError> {
+        if is_final {
+            data.extend_from_slice(&TRAILER);
```
erebe, line 421: Not really sure for this one.

But if possible it would be better to avoid the `extend_from_slice` as it can re-allocate/copy the whole vector to make more place.

If possible, it should be better to do an extra `decompress_vec` on error if `is_final` is set

src/extensions/compression/deflate.rs
```diff
@@ -0,0 +1,456 @@
...
+
+    fn peer_context_takeover(&self) -> bool {
+        match self.role {
+            Role::Server => !self.config.client_no_context_takeover,
+            Role::Client => !self.config.server_no_context_takeover,
+        }
+    }
+
+    // Compress the data of message.
+    pub(crate) fn compress(&mut self, data: &[u8]) -> Result<Vec<u8>, DeflateError> {
```
erebe, line 371: To be more aligned with the lib the signature should use `Bytes` instead of `&[u8]`

```rust
fn compress(&mut self, data: Bytes) -> Result<Bytes, DeflateError>
```


It would avoid turning the `Vec<u8>` into a `Bytes` when creating the Frame objet.
For the argument data, it does not change anything, it simply forward the Bytes buffer

src/extensions/compression/deflate.rs
```diff
@@ -0,0 +1,456 @@
...
+        output.truncate(output.len() - 4);
+
+        if !self.own_context_takeover() {
+            self.compressor.reset();
+        }
+
+        Ok(output)
+    }
+
+    pub(crate) fn decompress(
```
erebe, line 415: Same it should use `Bytes`, it would avoid conversion

src/extensions/compression/deflate.rs
```diff
@@ -0,0 +1,456 @@
...
+        let mut output = Vec::with_capacity(2 * data.len());
+        loop {
+            let offset = (self.decompressor.total_in() as usize) - before_in;
+            match self
+                .decompressor
+                .decompress_vec(&data[offset..], &mut output, FlushDecompress::None)
+                .map_err(|e| DeflateError::Decompress(e.into()))?
+            {
+                Status::Ok => output.reserve(2 * output.len()),
+                Status::BufError | Status::StreamEnd => break,
```
SvizelPritula, line 434: This doesn't check at all if the stream has actually ended with the final packed, I think.

akonradi-signal:
I'd like to request that the additions of `#[non_exhaustive]`, especially on enum types, be removed. For authors of libraries that use `tungstenite`, being able to rely on exhaustive matching lets us leverage the compiler to ensure at compile time that we've handled every possible error case appropriately. Without this guarantee, we have to include a branch for the "unknown" case in every `match` on an error so that there's a generic fallback path. The danger is that updating our tungstenite dependency from one version to another would change those generic fallback paths from never-used to sometimes-used. I'd much rather be forced, as a `tungstenite` consumer, to update my code when new error cases are added because my crate fails to compile.

Messages of doom aside, there's precedent already for having error cases that can't be produced without certain features being turned on. My library doesn't enable `tungstenite`'s TLS support, but we still have to handle the `Error::Tls` case. I'd rather do that (with an `unreachable!("tungstenite TLS isn't used")` than have to forever be vigilant for new error variants on upgrade because `#[non_exhaustive]` means the compiler is no longer helping me.

notonamap:
getting client-side support for this over line the in any form would be highly appreciated. we currently rely on patching in the signal fork via cargo which is kinda painful.

danielsufrindisler:
I currently have 2 different working branches for this.

### Branches
1.  A fork off of this branch with conflicts resolved. https://github.com/danielsufrindisler/tungstenite-rs/tree/permessage-deflate
2. A fork off of signalapp/tungstenite-rs:master with snapview/tungstenite-rs:master merged in https://github.com/danielsufrindisler/tungstenite-rs/tree/master

### Dependencies
Both versions include the headers crate

### Testing
The  signalapp version has several additional unit tests which are not present in this branch.

### Performance
There is similar performance for both branches (and master) with deflate disabled.

### Code
The folder structure is different between them.  There are a lot of additional helpful comments in the signalapp version.    

## Options
@daniel-abramov 
If either of these can be merged, great.  I'd recommend the signalapp version.  If you would prefer to move most of the new code into a different crate, I could do that as well.  If I'm not being helpful, or if I'm making things more complicated, I'll stop.

odedlaz:
@daniel-abramov @agalakhov

I want to pick this up. There're a bunch of companies, including [the one I work for](https://www.wsj.com/pro/cybersecurity/neo-raises-100-million-from-stealth-as-investors-back-battle-tested-cyber-execs-07e2780e), [signal](https://github.com/signalapp/tungstenite-rs), [openai](https://github.com/openai-oss-forks/tungstenite-rs)   that fork tungstenite only for this feature.

given the size & perf concerns in previous discussions I took a different angle while reading previous PRs that failed to get in as bedrock: #144, #235, #328 & this one.

I got this implemented in ~half LOCs s.t ~15% of tungstenite total LOCs are deflate. We’re using this in production and gain ~3x reduction in payload size. It passes all autobahn 216 permessage-deflate cases.

Do you prefer a PR to add PMD or an external crate? I have both tbh. I'll do what you ask to get this merged, just tell me what you prefer.

Thanks!

goriunov:
Agree, imo in majority of cases you need per message deflate and there is really no stable/maintained websocket library that supports deflate. For all the projects we build had to fork the tungstenite just to add compression.

PR #427, opened 2024-05-18
Set default-features = false for rustls

rukai:
Before releasing 0.22, I believe we should change how we depend on rustls.
Since rustls swapped to having multiple backends, we need to set `default-features = false` for rustls so that the user of tungstenite is free to choose the rustls backend.
This is particularly important since the new default backend for rustls does not build in many environments due to older version of gcc.
This PR needs to land before 0.22 since its a breaking change.

The `std` feature is enabled since it is [enabled by default](https://github.com/rustls/rustls/blob/4e2856b1fa4917a6a8c35c5c4aada1f5fd649cc5/rustls/Cargo.toml#L31) and required for tungstenite's rustls feature to compile at all

There is currently a PR open against tokio-tungstenite to do the same https://github.com/snapview/tokio-tungstenite/pull/332
But without this PR the tokio-tungstenite PR will have no effect.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,8 @@ version = "0.2.3"
 
 [dependencies.rustls]
 optional = true
+default-features = false
+features = ["std"]
 version = "0.23.0"
 
 [dependencies.rustls-pki-types]


daniel-abramov (approved):
Thanks! Unfortunately I've missed this one for some reason and I've just released 0.22.0 and noticed this one afterwards. I could publish 0.23.0 though.

rukai:
That's unfortunate, our project won't be able to upgrade to 0.22 as is. up to you whether to release 0.23 now or later. We can continue running off a git dependency for a while longer if needed.

daniel-abramov:
Done! I'm also going to publish `tokio-tungstenite` 0.23.0 soon.

PR #431, opened 2024-06-20
Fix handshake for `arti-client`

yukibtc:
As requirement for usage of `arti-client`, added a flush method call in the `HandshakeMachine` writing round.

### DIFF
--- a/src/handshake/machine.rs
+++ b/src/handshake/machine.rs
@@ -73,6 +73,7 @@ impl<Stream: Read + Write> HandshakeMachine<Stream> {
             HandshakeState::Writing(mut buf) => {
                 assert!(buf.has_remaining());
                 if let Some(size) = self.stream.write(Buf::chunk(&buf)).no_block()? {
+                    self.stream.flush()?;
                     assert!(size > 0);
                     buf.advance(size);
                     Ok(if buf.has_remaining() {


daniel-abramov (commented):
While this is a minimal change, I'm not sure about the possible implications. Could you explain why you think this change is necessary/essential?

From the description, it feels like it might be a patch for a particular issue within the aforementioned library, not with `tungstenite`.

yukibtc:
Yeah, the description is focused in particular for `arti-client` that uses `DataStream` instead of `TcpStream`. 

From the rust `std` lib docs, on the `Write::write` method: "This function will attempt to write the entire contents of buf, **but the entire write might not succeed**, or the write may also generate an error".
While the `Write::flush` method docs says: "Flush this output stream, **ensuring that all intermediately buffered contents reach their destination**."

So, in general, streams may require a flush to ensure that buffered contents reach their destination. Since `tungstenite` allow to use streams different from `TcpStream`, I think that a flush is required there, to allow to complete the handshake.

daniel-abramov:
> (..) on the Write::write method: "This function will attempt to write the entire contents of buf, **but the entire write might not succeed**, or the write may also generate an error" (..)

AFAIK the "might not succeed" refers to the "entire write", i.e. it says that there is no guarantee that the **whole buffer** will be written to the stream - in some cases only a part of what you pass will be written and the amount of bytes written is returned on a successful write.

However it's true that if there is any buffering on a stream, then if you want to guarantee that the data was actually sent, you must call `flush()`. So, if I understand correctly, you're saying that flushing after each write during the handshake process makes sense because the underlying stream may have some buffering and since we currently don't flush the stream **during the handshake,** this might pose an issue?

If so, do you think that `self.stream.flush()?;` is safe in that place (perhaps `self.stream.flush().no_block()?;` would be a better choice?).

yukibtc:
> So, if I understand correctly, you're saying that flushing after each write during the handshake process makes sense because the underlying stream may have some buffering and since we currently don't flush the stream during the handshake, this might pose an issue?

Yes, but I'm not sure if that is the best position where to flush. Maybe could be better to flush when the `HandshakeState::Writing` buf is empty?

daniel-abramov:
> > So, if I understand correctly, you're saying that flushing after each write during the handshake process makes sense because the underlying stream may have some buffering and since we currently don't flush the stream during the handshake, this might pose an issue?
> 
> Yes, but I'm not sure if that is the best position where to flush. Maybe could be better to flush when the `HandshakeState::Writing` buf is empty?

The place might be right, but then we might possibly need to handle a case when it blocks - what if `flush()` returns EWOULDBLOCK? - Ignoring that error does not seem to be a good idea as it would prevent from flushing the stream (the very reason of adding `flush()` at the first place).

In case of `write()`, the `no_block()` part was fine as we have an `else` block that essentially handles this error by returning our own wrapper type that is properly handled by the handshake logic.

daniel-abramov (changes requested):

daniel-abramov (approved):
LGTM!

I'll additionally ping @agalakhov to be on the safe side (he is the original author of the code). @agalakhov, some context: https://github.com/snapview/tungstenite-rs/pull/431#issuecomment-2286883789

agalakhov:
This looks Ok. I don't see why this wouldn't work. Maybe it changes the corner case behavior slightly, but it isn't a concern as long as all Autobahn tests are green.

agalakhov:
:+1:

yukibtc:
Approx. when will be released the new version on `crates.io`?

daniel-abramov:
Published! (for now only `tungstenite`)

PR #432, opened 2024-07-02
Updated dependencies

ivanrg99:
Updated some of the dependencies on the project

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ version = "0.26"
 
 [dev-dependencies]
 criterion = "0.5.0"
-env_logger = "0.11.5"
+env_logger = "0.11"
 input_buffer = "0.5.0"
 rand = "0.8.4"
 socket2 = "0.5.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,16 +29,16 @@ __rustls-tls = ["rustls", "rustls-pki-types"]
 
 [dependencies]
 data-encoding = { version = "2", optional = true }
-byteorder = "1.3.2"
-bytes = "1.0"
-http = { version = "1.0", optional = true }
-httparse = { version = "1.3.4", optional = true }
-log = "0.4.8"
-rand = "0.8.0"
+byteorder = "1.5"
+bytes = "1.6"
+http = { version = "1.1", optional = true }
+httparse = { version = "1.9", optional = true }
+log = "0.4"
+rand = "0.8"
 sha1 = { version = "0.10", optional = true }
-thiserror = "1.0.23"
-url = { version = "2.1.0", optional = true }
-utf-8 = "0.7.5"
+thiserror = "1.0"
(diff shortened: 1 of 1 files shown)


ivanrg99:
`env_logger` and other packages (like `syn`) will fail due to the MSRV being 1.60, which is quite old at this point. We should consider increasing the MSRV

daniel-abramov:
We once had to downgrade it https://github.com/snapview/tungstenite-rs/pull/432, but maybe it's not an issue anymore.

And yeah, I believe that this one https://github.com/snapview/tungstenite-rs/pull/411 fails exactly because of this.

ivanrg99:
You are totally right about the missing patch version. I though it was a shortcut for the latest patch version available, but I doubled check the docs and you are correct. I fixed it in my local branch.

As to bumping the MSRV from 1.60.0 to 1.63.0, this still poses problems. CI only runs a `cargo check` to see if everything is fine, but version 1.63 is unable to compile the tests (with all the dependencies being updated, that is). I have tried different combinations, and the lowest MSRV that I can get to pass clean (including running all tests using that version, using all possible feature combinations, and passing with all the latest dependencies) is 1.65.0. The tradeoff is that for 1.65.0 to be able to run the tests, some of the dev dependencies can't be upgraded. The only two affected are `criterion` (stuck in 0.3.6) and `env_logger` (stuck in 0.10.2). I think that just using a `cargo check` as a means to validate the MSRV is not that great, since we can't really run the tests to make sure if works. Ultimately it's up to whatever the project needs/decides. I just see that most of the Github Actions that check the MSRV fail, so maybe it's worth bumping that number of to 1.71.0

As an anecdote, I ran the benchmarks with all the dependencies updated to their latest versions, and I saw some performance regressions in my machine (M3 Max) on the `InputBuffer` and `ReadBuffer (heap)` benchmarks. I still think it is worth upgrading, since security and bug fixes/correctness should be the main priority, but still interesting to point out.

Let me know what you decide

daniel-abramov:
> You are totally right about the missing patch version. I though it was a shortcut for the latest patch version available, but I doubled check the docs and you are correct. I fixed it in my local branch.

Would you revert the aforementioned changes then? :) So that at least we have a bit cleaner diff with versions that were actually updated.

> I just see that most of the Github Actions that check the MSRV fail, so maybe it's worth bumping that number of to 1.71.0

Right, 1.71.0 is the MSRV of the `env_logger`. Truthfully, 1.71.0 is more than a year old, so it is a reasonable choice. However, I don't know how many projects would benefit from having a much lower MSRV. I don't have a strong opinion on MSRV, so I would be totally fine with 1.71.0. Unless anyone voices a strong concern against that, we could increase MSRV.

> As an anecdote, I ran the benchmarks with all the dependencies updated to their latest versions, and I saw some performance regressions in my machine (M3 Max) on the InputBuffer and ReadBuffer (heap) benchmarks.

How much though? Is it a significant difference?

Issue #312, opened 2022-11-04
Error's `Display` and `Error::source` contain duplicate info


vilgotf:
`Error` returns duplicate information since the error source is included in *both* the `Display` and `Error::source`, for example logging errors with tracing produce the following:
```
2022-11-04T19:05:01.777878Z  WARN crate: error=websocket errored error.sources=[IO error: unexpected end of file, unexpected end of file]
```

It would be nice to either include downstream error information in *just* the `Error::source` (leaving `Display` with "IO error") as discussed here: https://github.com/dtolnay/thiserror/issues/38 or to not implement `Error::source` for the `Error::Io` variant.

agalakhov:
Thank you for pointing this. This is one of artifacts coming from pre-2018 Rust and from early versions of Tungstenite. This definitely needs reworking.

PR #434, opened 2024-07-04
fix: eliminate linter errors and improve errors (#312)

tnychn:


agalakhov:
What's the point of converting abbreviated names like "TLS" or "HTTP" to lowercase?

daniel-abramov (commented):
I second the question - I ran `clippy` (nightly) locally and there were no warnings about the error message case (only a couple `result_large_err` warnings).

PS: Also, I believe that if we were to handle linter warnings in this PR, it would make sense to introduce a corresponding step in the CI pipeline.

daniel-abramov:
Superseded by https://github.com/snapview/tungstenite-rs/pull/445

PR #437, opened 2024-08-02
Fix behavior when connecting wss url without TLS support

maugier:
Hi,

  When compiled without TLS support, the client currently falls back to HTTP (and by default, on port 443) even if the URI has the `wss` scheme. This creates a problem that is hard to debug, as the call fails with a  generic WSS protocol error.

  If a `wss://` URI is requested and no TLS support is available, the client should fail fast with a `UrlError::TlsFeatureNotEnabled`.

Cheers,

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -52,6 +52,12 @@ pub fn connect_with_config<Req: IntoClientRequest>(
     ) -> Result<(WebSocket<MaybeTlsStream<TcpStream>>, Response)> {
         let uri = request.uri();
         let mode = uri_mode(uri)?;
+
+        #[cfg(not(any(feature = "native-tls", feature = "__rustls-tls")))]
+        if let Mode::Tls = mode {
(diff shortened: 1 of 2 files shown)


daniel-abramov (commented):
This looks like a useful thing! I think it would be even more awesome thing if we could cover it with a small test 🙂

maugier:
Writing the test made me realize that the failure should happen earlier, before we even attempt to connect the TCP stream.

I reverted the earlier patch and moved the check earlier. What do you think of this ?

daniel-abramov:
> I reverted the earlier patch and moved the check earlier. What do you think of this ?

Looks very sensible, thanks!

daniel-abramov (approved):

daniel-abramov:
So the only job that still fails does this only due to our low MSRV, as discussed in https://github.com/snapview/tungstenite-rs/pull/432 and other related issues, so I'll merge it and raise MSRV in a separate commit then.

PR #444, opened 2024-09-07
Add invalid request handling callback for websocket server

### LINKED ISSUE #443: Support custom behavior for some handshake error cases
I found that currently the server side of tungstenite will directly drop the connection when encountering some handshake failures (the `stream` is moved into `accept()` so it will be dropped when an error occurs).

However I have a requirement to return a valid http response when some websocket related headers are missing (such as the connection upgrade header). Currently tungstenite seems unable to do this.

I wonder if some new accept functions can be added to tungstenite to support this?

PragmaTwice:
It closes #443.

> I found that currently the server side of tungstenite will directly drop the connection when encountering some handshake failures (the stream is moved into accept() so it will be dropped when an error occurs).
>
> However I have a requirement to return a valid http response when some websocket related headers are missing (such as the connection upgrade header). Currently tungstenite seems unable to do this.

This patch provides a method to solve this problem.

A new method `on_invalid_request()` (with a default impl) is added into `Callback`. (so that no users code will be broken.)

By default the behavior has no change: the connection will just be dropped if the server get some invalid request (e.g. no connection upgrade header), but users can define their own `on_invalid_request()` to return a valid response for it.

It also makes the library more compliant to the standards, refer to RFC 6455:
> If the server, while reading the handshake, finds that the client did
   not send a handshake that matches the description below (note that as
   per [[RFC2616](https://datatracker.ietf.org/doc/html/rfc2616)], the order of the header fields is not important),
   including but not limited to any violations of the ABNF grammar
   specified for the components of the handshake, the server MUST stop
   processing the client's handshake and ***return an HTTP response*** with an
   appropriate error code (such as 400 Bad Request).


agalakhov:
An alternative solution would be returning `HandshakeError::Failure(Error, Connection)` so that the connection would be usable.  But the callback is fine too.

PR #445, opened 2024-09-08
Fix clippy warnings & add CI build job for clippy

nickelc:
This PR fixes the current default clippy warnings and some selected warnings from the `clippy::pedantic` group.

It also adds an additional CI build job for clippy as suggested in https://github.com/snapview/tungstenite-rs/pull/434#pullrequestreview-2235867363.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,16 @@ jobs:
           components: rustfmt
       - run: cargo fmt --all --check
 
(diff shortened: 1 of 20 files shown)


daniel-abramov (approved):
This is nice! Thanks for picking [it](https://github.com/snapview/tungstenite-rs/pull/434#pullrequestreview-2235867363) and improving!

PR #461, opened 2024-12-10
FEAT: Replace Vec<u8> with Bytes

Zarathustra2:
Hi I have given adding `Bytes` a try but I at some places I am not sure if there is a performance implication or not. This is a continuation of #104. This is the first time for me working with more low level network stuff so I might be missing one or two obvious things.

The breaking change here would be for users that use `Message::Binary())` instead of `Message::binary()` to construct the message. The fix for this is quite simple as they would have to just change it to `Message::Binary(data.into()))`. 

Let me know what you think. 

Thanks a lot and have a great day!

### DIFF
--- a/src/protocol/frame/frame.rs
+++ b/src/protocol/frame/frame.rs
@@ -1,4 +1,5 @@
 use byteorder::{NetworkEndian, ReadBytesExt};
+use bytes::{Bytes, BytesMut};
 use log::*;
 use std::{
     borrow::Cow,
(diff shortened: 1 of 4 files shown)


Zarathustra2:
Oh didn't realize someone already did the work in https://github.com/snapview/tungstenite-rs/pull/175 in a better job, so closing this in favor of that PR. My bad, should have taken a better look at the already open PRs, not sure how I didn't notice it

Issue #96, opened 2019-11-26
Allow sending of shared data


jean-airoldie:
I think that it would make sense that `Message` be instead:
```rust
pub enum Message {
    Text(Arc<String>),
    Binary(Arc<Vec<u8>>), // maybe taking `Bytes` from the `bytes` crate would be cleaner.
    Ping(Vec<u8>),
    Pong(Vec<u8>),
    Close(Option<CloseFrame<'static>>),
}
```
This is because the websocket transport cannot guarantee messages to not be lost (from what I understand). Thus someone who wants to prevent messages from being lost would keep a list of messages and only consider them "sent" when they are ack'd by the peer. Using a `Arc` would greatly reduce the cost of doing so since clone operations would be more efficient.

najamelan:
Hmm,

given your other issue, I understand your concern, but I think your usecase is rather rare. Thus putting everybodies message data always in an Arc will incur an overhead all the time for everybody that doesn't need this.

I would suggest that a connection loss is an exceptional and erroneous situation, in which generally it's more acceptable to have a performance cost. What I'm suggesting is that you regenerate the messages that haven't been acknowledged in this situation. This can happen concurrently to establishing a new connection.

That being said, maybe using a data type like `Bytes` instead of `Vec<u8>` would open up options for copyless scenarios, including this one? Not sure about the performance trade-offs with `Bytes`.

jean-airoldie:
> Thus putting everybodies message data always in an Arc will incur an overhead all the time for everybody that doesn't need this.

> I would suggest that a connection loss is an exceptional and erroneous situation, in which case it's generally acceptable to have a performance cost. What I'm suggesting is that you regenerate the messages that haven't been acknowledged in this situation.

In my experience, a TCP crash is only a matter of time, especially for long running connections (16bit checksum, etc.). Thus if you want reliable messaging you need to keep a copy of each message locally until it is ack'd. Currently, you need to clone a `Vec<u8>` each time to do so, which is a massive additional overhead just to have reliability. So while it is true that the crash itself is not part of the hot path, the mechanism to ensure reliability very much is.

> That being said, maybe using a data type like Bytes instead of Vec<u8> would open up options for copyless scenarios, including this one? Not sure about the performance trade-offs with Bytes.

[...]

PR #462, opened 2024-12-11
message: support `bytes::Bytes` as a payload type

daniel-abramov:
### Preface

Recently, there [has been another PR](https://github.com/snapview/tungstenite-rs/pull/461) trying to solve the long-standing https://github.com/snapview/tungstenite-rs/issues/96.

The issue [had already been addressed in another PR that has been there for a while](https://github.com/snapview/tungstenite-rs/pull/175). I've noticed that I even approved the PR a while ago. Yet, it has not been merged for reasons I can't remember (AFAIR, I was waiting for another approval from another maintainer(s)/active users, but somehow it went stalled). Now, the discussions on the PR and the issue are so long that it's hard to get a brief overview of the current state and blockers, if any.

Yesterday, I quickly went through https://github.com/snapview/tungstenite-rs/pull/175 and the discussions on the issue to see if I could come up with something that would give it at least some resolution (https://github.com/snapview/tungstenite-rs/pull/175 has conflicts with a master branch) and create a small and easy-to-review PR to see if it's something that we can merge rather quickly.

### TL;DR

~~Essentially, the only thing this one changes is that all binary messages are backed by `Bytes` instead of a `Vec<u8>`. The distinction between `Owned` and `Shared` is internal and invisible to the end user.~~ (see https://github.com/snapview/tungstenite-rs/pull/462#issuecomment-2536226200). I also have not touched the text messages for now (as there have been several proposals, and I'm not sure there is a consensus as to which one is the best).

~~So the question is if we are good with having `Bytes` instead of a `Vec<u8>` or if we want to expose that `Payload` type and make `Message` rely on it (like in #175). If I'm not mistaken, the cost of `Bytes` is negligible~~ (see https://github.com/snapview/tungstenite-rs/pull/462#issuecomment-2536226200); the main concern for using `Bytes` as the type for the `Message::Binary` was the additional allocation/copy of a buffer that is done before applying a mask (since `Bytes` is immutable), and #175 did so by creating a copy before applying a mask. I changed it so we don't make this copy if `Bytes` has unique access.

Does it make any sense, or did I miss something?

---

@Dushistov, @sdroege, @francois-random: I decided to mention you folks since you participated in the discussion the most in the past year and seem interested in the outcome of this one. I'm going to also mention @alexheretic (you seem to be interested in performance-related improvements 🙂).

-- 

Fixes https://github.com/snapview/tungstenite-rs/issues/96
Supersedes https://github.com/snapview/tungstenite-rs/pull/175 if we agree that it's a sensible solution.

### DIFF
--- a/src/protocol/frame/frame.rs
+++ b/src/protocol/frame/frame.rs
@@ -1,14 +1,17 @@
-use byteorder::{NetworkEndian, ReadBytesExt};
-use log::*;
 use std::{
     borrow::Cow,
     default::Default,
     fmt,
     io::{Cursor, ErrorKind, Read, Write},
+    mem,
     result::Result as StdResult,
     str::Utf8Error,
     string::{FromUtf8Error, String},
 };
+
+use byteorder::{NetworkEndian, ReadBytesExt};
+use bytes::Bytes;
+use log::*;
 
 use super::{
     coding::{CloseCode, Control, Data, OpCode},
@@ -203,11 +206,47 @@ impl FrameHeader {
     }
 }
 
+#[derive(Debug, Clone, Eq, PartialEq)]
+enum Payload {
+    Owned(Vec<u8>),
+    Shared(Bytes),
+}
+
+impl Payload {
+    pub fn as_slice(&self) -> &[u8] {
+        match self {
+            Payload::Owned(v) => v,
+            Payload::Shared(v) => v,
+        }
+    }
+
+    pub fn as_mut_slice(&mut self) -> &mut [u8] {
+        match self {
+            Payload::Owned(v) => &mut *v,
+            Payload::Shared(v) => {
+                // Using `Bytes::to_vec()` or `Vec::from(bytes.as_ref())` would mean making a copy.
+                // `Bytes::into()` would not make a copy if our `Bytes` instance is the only one.
+                let data = mem::take(v).into();
+                *self = Payload::Owned(data);
+                let Payload::Owned(v) = self else { unreachable!() };
+                v
+            }
+        }
+    }
+
+    pub fn into_data(self) -> Vec<u8> {
+        match self {
+            Payload::Owned(v) => v,
+            Payload::Shared(v) => v.into(),
+        }
+    }
+}
+
 /// A struct representing a WebSocket frame.
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct Frame {
     header: FrameHeader,
-    payload: Vec<u8>,
+    payload: Payload,
 }
 
 impl Frame {
@@ -215,7 +254,7 @@ impl Frame {
     /// This is the length of the header + the length of the payload.
     #[inline]
     pub fn len(&self) -> usize {
-        let length = self.payload.len();
+        let length = self.payload.as_slice().len();
         self.header.len(length as u64) + length
     }
 
@@ -239,14 +278,14 @@ impl Frame {
 
     /// Get a reference to the frame's payload.
     #[inline]
-    pub fn payload(&self) -> &Vec<u8> {
-        &self.payload
+    pub fn payload(&self) -> &[u8] {
+        self.payload.as_slice()
     }
 
     /// Get a mutable reference to the frame's payload.
     #[inline]
-    pub fn payload_mut(&mut self) -> &mut Vec<u8> {
-        &mut self.payload
+    pub fn payload_mut(&mut self) -> &mut [u8] {
+        self.payload.as_mut_slice()
     }
 
     /// Test whether the frame is masked.
@@ -269,36 +308,36 @@ impl Frame {
     #[inline]
     pub(crate) fn apply_mask(&mut self) {
         if let Some(mask) = self.header.mask.take() {
-            apply_mask(&mut self.payload, mask);
+            apply_mask(self.payload.as_mut_slice(), mask);
         }
     }
 
     /// Consume the frame into its payload as binary.
     #[inline]
     pub fn into_data(self) -> Vec<u8> {
-        self.payload
+        self.payload.into_data()
     }
 
     /// Consume the frame into its payload as string.
     #[inline]
     pub fn into_string(self) -> StdResult<String, FromUtf8Error> {
-        String::from_utf8(self.payload)
+        String::from_utf8(self.payload.into_data())
     }
 
     /// Get frame payload as `&str`.
     #[inline]
     pub fn to_text(&self) -> Result<&str, Utf8Error> {
-        std::str::from_utf8(&self.payload)
+        std::str::from_utf8(self.payload.as_slice())
     }
 
     /// Consume the frame into a closing frame.
     #[inline]
     pub(crate) fn into_close(self) -> Result<Option<CloseFrame<'static>>> {
-        match self.payload.len() {
+        match self.payload.as_slice().len() {
             0 => Ok(None),
             1 => Err(Error::Protocol(ProtocolError::InvalidCloseSequence)),
             _ => {
-                let mut data = self.payload;
+                let mut data = self.payload.into_data();
                 let code = u16::from_be_bytes([data[0], data[1]]).into();
                 data.drain(0..2);
                 let text = String::from_utf8(data)?;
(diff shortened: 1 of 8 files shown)


src/protocol/frame/frame.rs
```diff
@@ -203,19 +206,55 @@ impl FrameHeader {
     }
 }
 
+#[derive(Debug, Clone, Eq, PartialEq)]
+enum Payload {
```
Dushistov, line 210: Not sure that follow, but  can be this new type be replaced by 

[Bytes::is_unique](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.is_unique) and `Bytes::to_vec` ?

If no clone of `Bytes` create from `Vec`, than `is_unique` should gives `true`.

Zarathustra2:
I am not sure if this is just on my end but when I run `cargo bench` on master I get:
```
     Running benches/write.rs (target/release/deps/write-9e4d5d42fd6d53b9)
Gnuplot not found, using plotters backend
                        time:   [5.3589 ms 5.3623 ms 5.3665 ms]
                        change: [-0.1032% +0.0279% +0.1456%] (p = 0.67 > 0.05)
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) low mild
  1 (1.00%) high mild
  2 (2.00%) high severe
```

and when running for this PR/branch:
```
     Running benches/write.rs (target/release/deps/write-9e4d5d42fd6d53b9)
Gnuplot not found, using plotters backend
                        time:   [7.2634 ms 7.2756 ms 7.2928 ms]
                        change: [+35.427% +35.679% +36.006%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
```

daniel-abramov:
Thanks, benchmarking is a good thing! I'll investigate what's going on.

src/protocol/frame/frame.rs
daniel-abramov: Do you mean that instead of a separate `Payload` type, we could use `Bytes`? (note: this is a payload of a frame, not the `Message`; the `Message` uses `Bytes`).

Using `Bytes` as a type of the `payload` of a `Frame` would be possible if not masking that we need to perform when e.g. sending frames from the client. `Bytes` is immutable. `BytesMut` is, though. But converting it to `BytesMut` upfront (before we know that we'll mutate the payload by applying a mask) is probably not optimal (I hope I understood your question correctly).

alexheretic:
I think this is a step in the right direction.

Nice to see the write bench was useful :) I think we would benefit from some more scenarios benched in this way, binary writes & reads etc. If I can find some time I'll raise a PR proposing some more.

agalakhov:
@alexheretic IIRC, I looked at the `bytes` crate while writing the very first version of Tungstenite, and the only reason I did not take it was the API instability at that time. As `bytes` is a solid part of the ecosystem now, it's a good idea to use it.

If implemented correctly, converting to `Bytes` should not affect benchmarks at all, being of true zero cost. It is only important to make correct choices between `Bytes` and `Buf`/`BufMut` at all places. I believe that it is even possible to _improve_ performance compared to the old approach, as scatter-gather I/O is now stabilized in `std`. Using `read_vectored()`/`write_vectored()` together with a collection of `BufMut`/`Bytes` could make Tungstenite to true _zero-copy_.

alexheretic:
I noticed you can use this branch for writing text too with `Frame::message(str_bytes, OpCode::Data(Data::Text), true)`

I struggled to produce a benchmark win using that though, even though it should really eliminate allocation. Maybe allocation just isn't significant enough to the other write work currently?

I raised a PR with the extra benches: https://github.com/snapview/tungstenite-rs/pull/463

daniel-abramov:
> I struggled to produce a benchmark win using that though, even though it should really eliminate allocation. Maybe allocation just isn't significant enough to the other write work currently?

I don't expect any improvements in benchmarks from this PR, at least for the benches in question. The reason is that this change does not really switch to `Bytes` completely (I've just updated the PR title because it referred to the first attempt to just switch to `Bytes`). The TL;DR is that it switches from `Vec<u8>` as a payload and message type to a `Payload` type that is defined as:

```
pub enum Payload {
    /// Owned data with unique ownership.
    Owned(Vec<u8>),
    /// Shared data with shared ownership.
    Shared(Bytes),
}
```

In other words, folks that used `tungstenite` with owned data (`String` or `Vec<u8>`) should not notice any difference at all because their messages are internally converted to `Payload::Owned(Vec<u8>)`. But those who wanted to use `Bytes`, are now capable of doing it as well. This means that the only performance improvement I expect from this kind of change is that the users who write servers (specifically those who broadcast messages to many recipients or send the same message several times for whatever reasons).

That's why there was no difference in performance when I compared it to a master branch (the benchmarks in master use owned data).

However, there is a difference when testing it against your new benchmarks. This PR version shows, on average, a 4-5% decrease in performance in reading and writing. I'm still unsure what this could be attributed to, given that the underlying type for all benchmarks remains the same even for this PR (no `Bytes` involved, just `Vec<u8>`). My only [non-scientific] guess is that the machine representation of a `Payload` is a couple of bytes larger. This is probably invisible for regular use, but for a benchmark that sends 100k messages, this tiny cost might accumulate and result in something visible as a minor performance decrease (after all, all messages in the benchmark are pretty small).

I then modified your writing benchmark for the use cases explicitly mentioned in https://github.com/snapview/tungstenite-rs/issues/96. Instead of creating small messages in a loop, I made a moderately sized message (1KB) and sent it in that loop in the writing benchmark. I ran it on this branch and then compared it to the version that uses `Vec<u8>` in master, and the performance increase is quite massive:

```
                        time:   [8.3416 ms 8.3952 ms 8.4669 ms]
                        change: [+123.71% +126.18% +128.52%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 10 outliers among 100 measurements (10.00%)
  2 (2.00%) low severe
  7 (7.00%) low mild
  1 (1.00%) high severe
```

So the master version is **126% slower** than the version that uses `Bytes` (due to the massive amount of allocations caused by constant cloning).

The net result of the changes is that performance for general use cases drops by about 5% (and I'm not sure why, apart from the guess mentioned above). However, for server users, the performance wins may be above 125% if they specifically choose to use `Bytes` for cases when it's appropriate (broadcasting the same message, etc) since they avoid all of the costs associated with expensive cloning.

Tbh, if that small performance drop for the general use case could somehow be eliminated, then switching to the new `Payload` type would be a no-brainer.

agalakhov:
In fact, `Vec` is not needed here at all, as it could be replaced as well.

alexheretic:
fwiw I didn't see much difference between the branches in the new benches. Though I did convert them to this branch slightly differently and I think I stumbled upon [`slice::to_vec` being faster](https://github.com/snapview/tungstenite-rs/pull/464) which skewed the comparison in this branch's favour.

Now I run using `to_vec` in both branches I see the following, which I interpret as 
* Read speed is about the same
* Write speed is slightly impacted
```
                        time:   [13.010 ms 13.028 ms 13.101 ms]
                        change: [+2.1102% +2.8253% +3.5454%] (p = 0.10 > 0.05)

                        time:   [11.052 ms 11.053 ms 11.053 ms]
                        change: [+2.0613% +2.2854% +2.5105%] (p = 0.08 > 0.05)

                        time:   [4.6399 ms 4.6452 ms 4.6661 ms]
                        change: [+11.700% +12.027% +12.354%] (p = 0.09 > 0.05)

                        time:   [5.6127 ms 5.6169 ms 5.6338 ms]
                        change: [+3.1398% +3.6457% +4.1546%] (p = 0.10 > 0.05)
```

daniel-abramov:
Thanks! Yep, this roughly aligns with the results that I'm getting.

> I then modified your writing benchmark for the use cases explicitly mentioned in https://github.com/snapview/tungstenite-rs/issues/96. Instead of creating small messages in a loop, I made a moderately sized message (1KB) and sent it in that loop in the writing benchmark. I ran it on this branch and then compared it to the version that uses Vec<u8> in master, and the performance increase is quite massive:

I'm thinking about adding this benchmark. This means that the version in `master` would then create a single message and send it multiple times (cloning it each before sending), while the version from the branch would wrap it in `Bytes` and pass it to `tungstenite` (eliminating the need to do expensive cloning).

alexheretic:
tbh I don't see the minor benches perf regression as blocking this PR. 

The scenario where a server broadcasts a message to N clients is clearly a massive advantage to supporting Bytes. The regression reported by the benches is not large.

alexheretic:
I got carried away a bit trying to adding text support :sweat_smile: https://github.com/snapview/tungstenite-rs/pull/465

daniel-abramov:
Awesome! Since yours is rebased on top of this branch, I'll merge it instead of integrating these two separately! So far, it seems like no one objected, so I think we're good to go!

PR #464, opened 2024-12-13
Bench: optimise binary data generation

alexheretic:
I noticed `slice::to_vec` seems to be faster than `.into()` :shrug:. This code is also more compatible with the proposed approach in #462 

As this generates vec data faster it makes the bench faster (but just the bench) .

```
                        time:   [4.0320 ms 4.0390 ms 4.0672 ms]
                        change: [-8.8302% -8.3994% -7.9684%] (p = 0.05 > 0.05)

                        time:   [5.3872 ms 5.3921 ms 5.4115 ms]
                        change: [-8.0997% -7.6006% -7.0983%] (p = 0.07 > 0.05)
```

### DIFF
--- a/benches/read.rs
+++ b/benches/read.rs
@@ -51,7 +51,7 @@ fn benchmark(c: &mut Criterion) {
                 for i in 0_u64..100_000 {
                     writer
                         .send(match i {
-                            _ if i % 3 == 0 => Message::Binary(i.to_le_bytes().into()),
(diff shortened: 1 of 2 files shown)


daniel-abramov (approved):
> I noticed slice::to_vec seems to be faster than .into() 🤷

Huh! That's very interesting. Would not expect that either!

PR #465, opened 2024-12-14
Support `bytes::Bytes` & `bytes::BytesMut` payloads for binary & text messaging

alexheretic:
This PR builds on top of #462 
* Adds `Utf8Payload` for text messages backed by `Payload`.
* Tri-variant `Payload` for `Bytes`, `BytesMut` & `Vec`.
  - `Bytes` gives flexibility supporting shared data & cheap cloning.
  - `BytesMut` provides best read performance as we can get these from the new read buffer sans copying & unmask optimially too.
  - `Vec` provides best write performance. I couldn't get `BytesMut` to perform as well even changing the bench user code. Since it doesn't increase Payload 40 bytes size it seems worth having.
* Replace the read buffer with a `BytesMut` based one that can then emit payloads without copying.
* Optimise incomplete message handling so complete messages do not need to be copied (general case for small messages).
* Add `Payload::share` for cheap cloning.
* Make `Payload::eq` compare data, so allow equality across variants.
* Remove and change a bunch of fns & types to suit the new internals.
* Add general &[u8]/&str From impl and use `const fn from_static` to construct static `Payload`/`Utf8Payload` data.

The main advantage here is extending the _bytes_ flexibility to Text types and better read performance.

### Benchmarks
Write performance remains a little slower, no more than [noted in the parent PR](https://github.com/snapview/tungstenite-rs/pull/462#issuecomment-2542530496). However, read performance is significantly faster.
```
$ cargo bench --bench \* -- --quick --noplot --baseline master
                        time:   [7.8768 ms 7.9103 ms 7.9187 ms]
                        change: [-38.736% -37.861% -36.967%] (p = 0.06 > 0.05)

                        time:   [7.1168 ms 7.1322 ms 7.1939 ms]
                        change: [-34.273% -33.840% -33.406%] (p = 0.07 > 0.05)

                        time:   [4.6548 ms 4.6852 ms 4.6928 ms]
                        change: [+6.0437% +6.5546% +7.0663%] (p = 0.10 > 0.05)

                        time:   [6.0834 ms 6.0875 ms 6.1039 ms]
                        change: [+6.0147% +6.6040% +7.1980%] (p = 0.10 > 0.05)
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ __rustls-tls = ["rustls", "rustls-pki-types"]
 [dependencies]
 data-encoding = { version = "2", optional = true }
 byteorder = "1.3.2"
-bytes = "1.0"
+bytes = "1.9.0"
 http = { version = "1.0", optional = true }
 httparse = { version = "1.3.4", optional = true }
 log = "0.4.8"
--- a/benches/read.rs
+++ b/benches/read.rs
@@ -52,7 +52,7 @@ fn benchmark(c: &mut Criterion) {
                     writer
                         .send(match i {
                             _ if i % 3 == 0 => Message::binary(i.to_le_bytes().to_vec()),
-                            _ => Message::Text(format!("{{\"id\":{i}}}")),
+                            _ => Message::text(format!("{{\"id\":{i}}}")),
                         })
                         .unwrap();
                     sum += i;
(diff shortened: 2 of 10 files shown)


daniel-abramov (approved):
You're fast! 🙂 

Interesting decision to switch from `ReadBuffer` to `BytesMut` for the `in_buffer`! It's nice that it helps to spare additional allocations (like that one with a vector that had been previously allocated just to read and return a payload from the in_buffer).

Zarathustra2:
Wow that is awesome!

@daniel-abramov Do you have an ETA when you plan on releasing 0.25 for tungstenite & tokio-tungstenite ?

daniel-abramov:
> Wow that is awesome!
> 
> @daniel-abramov Do you have an ETA when you plan on releasing 0.25 for tungstenite & tokio-tungstenite ?

Published! 🎉

PR #466, opened 2024-12-14
tiny optimization: use more optimized method for Vec<u8> to Bytes conversation

Dushistov:
Bytes::from_owner is "generic" method, for any "owners" of [u8], while specialized "impl From<Vec<u8>> for Bytes" can be faster in some cases, for example if Vec::capacity==Vec::len, then it is almost no-op.

### DIFF
--- a/src/protocol/frame/payload.rs
+++ b/src/protocol/frame/payload.rs
@@ -141,7 +141,7 @@ impl Payload {
                 *self = Self::Shared(mem::take(data).freeze());
             }
             Self::Vec(data) => {
-                *self = Self::Shared(Bytes::from_owner(mem::take(data)));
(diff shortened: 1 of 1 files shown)


Dushistov:
For context in which it was created: `Vec<u8>` -> `Bytes` and then clone,
the usage of `impl From<Vec<u8>> for Bytes` instead of `Bytes::from_owner` is not so important,

because of immediate clone, which add extra memory allocations which was omitted during `Bytes::from`,
but later because of usage `Vtable` for `Vec<u8>` it can provide benefits, like conversation to `Vec<u8>` if usage count is 1.

alexheretic (approved):

alexheretic:
lgtm thanks for the quick improvement :+1:

PR #471, opened 2024-12-17
Use set-len when reading into buffer

alexheretic:
Minor optimisation of read buffer logic using `set_len` to avoid writing zeroes, more worthwhile if configuring large read buffer sizes.

Also simplified the code a bit using the buffer directly instead of split_off + unsplit.

### DIFF
--- a/src/protocol/frame/mod.rs
+++ b/src/protocol/frame/mod.rs
@@ -19,7 +19,7 @@ use crate::{
 };
 use bytes::BytesMut;
 use log::*;
-use std::io::{Cursor, Error as IoError, ErrorKind as IoErrorKind, Read, Write};
+use std::io::{self, Cursor, Error as IoError, ErrorKind as IoErrorKind, Read, Write};
 
 /// Read buffer size used for `FrameSocket`.
 const READ_BUF_LEN: usize = 128 * 1024;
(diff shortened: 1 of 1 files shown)


daniel-abramov (approved):

daniel-abramov:
Thanks for all performance-related improvements, @alexheretic! Once this one is merged, we'll be good to go with the release, right? (you don't plan further optimizations for now, do you? 🙂 )

alexheretic:
> Thanks for all performance-related improvements, @alexheretic! Once this one is merged, we'll be good to go with the release, right? (you don't plan further optimizations for now, do you? 🙂 )

I don't have anything queued up no :sweat_smile:  so if you're happy with all the API changes then a new release is fine by me :+1:

conradludgate (commented):
I'm pretty sure this is unsound. A dodgy but valid Read::read impl is able to read the uninitialised bytes.

daniel-abramov:
Great point! I also have not thought about that until now.

> Implementations of this method **can make no assumptions about the contents** of buf when this function is called. It is recommended that implementations only write data to buf instead of reading its contents.

But at the same time it explicitly says:

> Correspondingly, however, callers of this method in unsafe code must not assume any guarantees about how the implementation uses buf. The trait is safe to implement, so it is possible that the code that’s supposed to write to the buffer might also read from it. It is your responsibility to make sure that buf is initialized before calling read. **Calling read with an uninitialized buf** (of the kind one obtains via [MaybeUninit<T>](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html)) **is not safe, and can lead to undefined behavior**.

---

@alexheretic, I assume I can release a new patch version with just reverted change (the previous implementation was sound and not much slower anyway as you mentioned)

alexheretic:
I'd rather remove just the set_len usage than revert the whole pr

daniel-abramov:
Makes sense!

**NB:** I've already published 0.26.1 with the commit reverted (I just immediately followed my reflexes and reverted the commit "just to be on the safe side" right away so that we have a safe published version before someone starts actively using 0.26.0)

PR #473, opened 2024-12-17
Rework as `Bytes`-only payload

alexheretic:
Spurred by the chat in https://github.com/tokio-rs/axum/pull/3078 This PR experiments with simplifying the API to use only `Bytes` as payload. It does work pretty well, ~the only downside seems to be client->server write performance as masking requires mutating the bytes.~ Now fixed.

* Read unmask performance was addressed by unmasking before converting to `Bytes` to avoid the overhead of having to mutate them.
* Write masking performance was addressed by writing and unmasking in the `out_buffer` rather than in-place.


alexheretic:
I wonder if there is a way of optimising the write to not need to convert `Bytes` too...

agalakhov:
Performance regression at 100k small messages seems to be significant. This however does not mean that it will be so in real use. I would test it a little bit more in other environments. The real question is, how many `Arc`s exist in real life and how many of them do exist only due to `Bytes`.

Such a performance impact (10%) in a real-life high load example would be huge. Ten percent more load means ten percent more servers in the cluster and ten percent more hardware costs, and hardware is expensive.

alexheretic:
Actually I found we don't _need_ to apply the mask then write, we have an `out_buffer` so we can write into the out_buffer _then_ apply the mask to the written slice. This has no extra cost and means we have no need to mutate the payloads at any point.

alexheretic:
Write performance looks good now

```
                        time:   [4.7412 ms 4.7432 ms 4.7437 ms]
                        change: [+2.1394% +2.4145% +2.6909%] (p = 0.08 > 0.05)

                        time:   [5.9248 ms 5.9282 ms 5.9416 ms]
                        change: [-1.9907% -1.5423% -1.0910%] (p = 0.05 < 0.05)
```

src/protocol/frame/frame.rs
```diff
@@ -304,7 +289,7 @@ impl Frame {
             0 => Ok(None),
             1 => Err(Error::Protocol(ProtocolError::InvalidCloseSequence)),
             _ => {
-                let mut data = self.payload.as_slice();
+                let mut data = BytesMut::from(self.payload);
```
alexheretic, line 292: This isn't ideal (though no worse than before), maybe we should also convert `CloseFrame::reason` to bytes?

alexheretic: I made that change

adryzz (approved):

PR #477, opened 2024-12-23
Remove byteorder, use bytes::Buf::get_uint instead

nylonicious:
`byteorder` was only used in one place and since bytes crate implements similar functionality it lets us drop it pretty easily.

### DIFF
--- a/src/protocol/frame/frame.rs
+++ b/src/protocol/frame/frame.rs
@@ -1,4 +1,3 @@
-use byteorder::{NetworkEndian, ReadBytesExt};
 use log::*;
 use std::{
     default::Default,
@@ -18,7 +17,7 @@ use crate::{
     error::{Error, ProtocolError, Result},
     protocol::frame::Utf8Bytes,
 };
-use bytes::{Bytes, BytesMut};
+use bytes::{Buf, Bytes, BytesMut};
 
 /// A struct representing the close command.
 #[derive(Debug, Clone, Eq, PartialEq)]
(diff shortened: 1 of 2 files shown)


hkBst (changes requested):
Removing byteorder is good, since it does not add much that is not already in std, but using the bytes crate here is also unnecessary. If you go explore [there](https://docs.rs/bytes/latest/src/bytes/buf/buf_impl.rs.html#58-65) you'll find it does:
```rust
let slice_at = match 8.checked_sub(length_length) {
    Some(slice_at) => slice_at,
    None => panic_does_not_fit(8, length_length),
};

let mut buf = [0; 8];
buffer[..length_length].as_ref().copy_to_slice(&mut buf[slice_at..]);
return u64::from_be_bytes(buf);
```
If we copy the bytes into our own 8-byte buffer but start at `slice_at` like the code above, then the copy is just not necessary.

nylonicious:
Thanks for pointing that out, hopefully I managed to get it right this time.

src/protocol/frame/frame.rs
```diff
@@ -161,14 +160,18 @@ impl FrameHeader {
...
             if length_length > 0 {
-                match cursor.read_uint::<NetworkEndian>(length_length) {
-                    Err(ref err) if err.kind() == ErrorKind::UnexpectedEof => {
-                        return Ok(None);
+                const SIZE: usize = mem::size_of::<u64>();
+                let mut buffer = [0; SIZE];
+                let start = match SIZE.checked_sub(length_length) {
+                    Some(start) => start,
+                    None => {
+                        panic!("the integer can fit {} bytes, but {} is given", SIZE, length_length)
```
Dushistov, line 168: Nit pick.

Wouldn't it be better to use the `unreachable!` macros here?

It is the same as `panic!` in terms of generated code, but better clarify that `websocket` length max is 8 bytes, exactly `sizeof::<u64>`, so it is impossible condition.

And even with rustc 1.63.0, it is possible to write:

```
               const fn extra_bytes(&self) -> usize {...}
                const SIZE: usize = LengthFormat::U64.extra_bytes();
```

so we can guarantee that `u64::from_be_bytes` can handle max websocket packet length,
and replace `panic!` with ` std::intrinsics::unreachable`, to make code faster.
But I suppose it is overkill to use `unsafe` for such case.

hkBst: An `assert!(length_length <= SIZE, panic_msg);` would let you use normal subtraction and is more direct, I think. Maybe that would express the unreachability enough?

nylonicious: I think without unsafe this gives us the best assembly, but I'm pretty bad at reading it.

hkBst (approved):

PR #478, opened 2025-01-13
upd: modifying websocket data packets to use layer8

muse254:
CHECKLIST:
- [x] Initial renaming
- [ ] Modify the client and server operations to use layer8 data encryption
- [ ] Create sample application that consumes the package on interceptor-rs


muse254:
Wrong repo

PR #484, opened 2025-02-18
Hash consistency for Borrow + Hash traits for Utf8Bytes

yhx-12243:
As mentioned in documentation of rust standard library, if a type implements both [`Borrow`](https://doc.rust-lang.org/nightly/std/borrow/trait.Borrow.html) and [`Hash`](https://doc.rust-lang.org/nightly/std/hash/trait.Hash.html), they should be consistent: `hash(x) == hash(x.borrow())`, (cf. https://rust-lang.github.io/rust-clippy/master/index.html#impl_hash_borrow_with_str_and_bytes)

However, the current implementation is not consistent. It may cause unexpected behaviors when using HashMap for these.

### DIFF
--- a/src/protocol/frame/utf8.rs
+++ b/src/protocol/frame/utf8.rs
@@ -3,7 +3,7 @@ use core::str;
 use std::fmt::Display;
 
 /// Utf8 payload.
-#[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
+#[derive(Debug, Default, Clone, Eq, PartialEq)]
 pub struct Utf8Bytes(Bytes);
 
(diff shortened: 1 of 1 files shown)


daniel-abramov (approved):
Good catch, thanks!

> (cf. https://rust-lang.github.io/rust-clippy/master/index.html#impl_hash_borrow_with_str_and_bytes)

NB: Note that the particular lint covers a _similar, but different_ case.

---

@sdroege, mentioning you just in case you rely on the current `Hash` implementation somewhere in other crates 🙂

sdroege:
I don't, thanks for catching this :)

PR #488, opened 2025-03-26
Add more details for utf8 error in order to debug properly

bnjjj:
We recently got an issue with a non utf8 compliant value in a specific header. We had no way to debug it properly. I had to fork tungstenite and add more details to know what was wrong. This PR intends to give more details and help when debugging that kind of issue.


daniel-abramov (approved):
Thanks for the change!

My only question: should the error handling not be consistent with the rest of the places? I.e. in `handshake/client.rs` the error is mapped to provide additional context to the error. While the rest of the places remain unchanged.

Not a big deal though - still better than no information at all!

PR #496, opened 2025-05-16
Fix large message read performance by enforcing max `read_buffer_size` read chunks

### LINKED ISSUE #493: Performance regression (compared to `0.24.0`) when reading large (100 MB+) messages in `0.25.0+`
With version 0.24 there where no problems in getting responses which where 50mb and bigger in the range of 900mb+ .

My question is how I can best debug or check which part could cause the problem that, it startet to make problems.
Before it could download with ~33mb/s from the websocket connection and now it looks like it does nothing.

Where could I best start to look which part could be the problematic part.

I have tested it with two different  creates https://github.com/gbaranski/ezsockets and https://github.com/jgraef/reqwest-websocket/ which both work if I use version 0.24 but both stop working if I go to the newest version.

When you want I can also test it directly with https://github.com/snapview/tokio-tungstenite if there the problem is also if needed.

alexheretic:
Non-breaking change to use  `read_buffer_size`  as a maximum read buffer size, so a large message will be read in chunks of, by default, 128KiB. This seems to significantly improves the performance of large messages.

See https://github.com/snapview/tungstenite-rs/issues/493#issuecomment-2885481091

Resolves #493

## Benchmarks
Using the #493 provided benches this fix addresses the regression and indeed provides better performance for all sizes than 0.24.

```
# 0.24
>>> 1.0 MB took ~1.993457ms with speed ~501.6 MB/s
>>> 3.0 MB took ~4.074073ms with speed ~736.4 MB/s
>>> 6.0 MB took ~5.40309ms with speed ~1.1 GB/s
>>> 10.0 MB took ~7.491196ms with speed ~1.3 GB/s
>>> 20.0 MB took ~16.038946ms with speed ~1.2 GB/s
>>> 30.0 MB took ~24.521725ms with speed ~1.2 GB/s
>>> 40.0 MB took ~37.509191ms with speed ~1.0 GB/s
>>> 50.0 MB took ~45.244393ms with speed ~1.1 GB/s
>>> 60.0 MB took ~53.531487ms with speed ~1.1 GB/s
>>> 100.0 MB took ~102.38638ms with speed ~976.7 MB/s
>>> 1000.0 MB took ~3.816925448s with speed ~262.0 MB/s

# 0.26.2
>>> 1.0 MB took ~1.402537ms with speed ~713.0 MB/s
>>> 3.0 MB took ~2.322722ms with speed ~1.3 GB/s
>>> 6.0 MB took ~3.837006ms with speed ~1.5 GB/s
>>> 10.0 MB took ~3.960994ms with speed ~2.5 GB/s
>>> 20.0 MB took ~31.777352ms with speed ~629.4 MB/s
>>> 30.0 MB took ~72.20983ms with speed ~415.5 MB/s
>>> 40.0 MB took ~95.672327ms with speed ~418.1 MB/s
>>> 50.0 MB took ~177.217286ms with speed ~282.1 MB/s
>>> 60.0 MB took ~261.481951ms with speed ~229.5 MB/s
>>> 100.0 MB took ~647.759606ms with speed ~154.4 MB/s
>>> 1000.0 MB took ~45.621549647s with speed ~21.9 MB/s

# PR
>>> 1.0 MB took ~550.883µs with speed ~1.8 GB/s
>>> 3.0 MB took ~1.555067ms with speed ~1.9 GB/s
>>> 6.0 MB took ~1.648007ms with speed ~3.6 GB/s
>>> 10.0 MB took ~2.242798ms with speed ~4.4 GB/s
>>> 20.0 MB took ~5.894016ms with speed ~3.3 GB/s
>>> 30.0 MB took ~11.865491ms with speed ~2.5 GB/s
>>> 40.0 MB took ~12.758398ms with speed ~3.1 GB/s
>>> 50.0 MB took ~21.80129ms with speed ~2.2 GB/s
>>> 60.0 MB took ~25.910093ms with speed ~2.3 GB/s
>>> 100.0 MB took ~44.506567ms with speed ~2.2 GB/s
>>> 1000.0 MB took ~635.858695ms with speed ~1.5 GB/s
```

I think we should add some high quality benches into this repo so we can track general performance better with a real `Read` impl. The current read benches are quite specific to batched small writes/reads. If I get time I'd like to add some as a follow up.


src/protocol/frame/frame.rs
```diff
@@ -66,6 +66,10 @@ impl Default for FrameHeader {
 }
 
 impl FrameHeader {
+    /// > longest possible header is 14 bytes, which would represent a message sent from
+    /// > the client to the server with a payload greater then 64KB.
+    pub(crate) const MAX_SIZE: usize = 14;
```
daniel-abramov, line 71:
```suggestion
    /// > The longest possible header is 14 bytes, which would represent a message sent from
    /// > the client to the server with a payload greater than 64KB.
    pub(crate) const MAX_SIZE: usize = 14;
```

src/protocol/frame/mod.rs
```diff
@@ -192,7 +196,14 @@ impl FrameCodec {
...
             // Not enough data in buffer.
-            self.in_buffer.reserve(self.header.as_ref().map(|(_, l)| *l as usize).unwrap_or(6));
+            if let Some((_, len)) = &self.header {
+                if !reserved_full_msg_len {
+                    self.in_buffer.reserve(*len as usize);
+                    reserved_full_msg_len = true;
+                }
+            } else {
+                self.in_buffer.reserve(FrameHeader::MAX_SIZE);
+            }
```
daniel-abramov, line 206:
```suggestion
            let reserve_len = (&self.header).map(|(_, len)| *len as usize).unwrap_or(FrameHeader::MAX_SIZE);
            self.in_buffer.reserve(reserve_len);
```

What do you think of this version? It helps to eliminate `let reserved_full_msg_len`; also, `reserve()` would not re-reserve memory if already reserved, if I'm not mistaken.

alexheretic: That's what we had before-ish. It makes sense for messages we'll read in a single chunk.

For large messages though, say 1MB, it will reserve 1MB after each `read_buffer_size`, say 128KiB, chunk read. So we'd end up over-reserving by up to 1MB.

We _could_ only reserve what we need for this chunk, which would solve that. But it seemed more efficient to reserve, and therefore allocate, once instead of multiple times per-chunk of a large message.

That's why I added the flag. For all messages we need only reserve the body len once. For large message we may loop multiple times calling `read_in`.

alexheretic:
With the proposed e2e benchmarks in #497 we also see the performance addressed there:

```
send+recv 1 KiB         time:   [13.309 µs 13.356 µs 13.546 µs]
                        change: [−2.0608% −0.7296% +0.6140%] (p = 0.54 > 0.05)
                        No change in performance detected.

send+recv 2 KiB         time:   [14.117 µs 14.225 µs 14.657 µs]
                        change: [−7.6625% −3.7156% +0.4184%] (p = 0.20 > 0.05)
                        No change in performance detected.

send+recv 4 KiB         time:   [15.365 µs 15.438 µs 15.731 µs]
                        change: [+0.8485% +2.2006% +3.5568%] (p = 0.11 > 0.05)
                        No change in performance detected.

send+recv 8 KiB         time:   [16.583 µs 16.635 µs 16.843 µs]
                        change: [−6.9808% −4.5915% −2.1161%] (p = 0.10 > 0.05)
                        No change in performance detected.

send+recv 16 KiB        time:   [19.908 µs 19.910 µs 19.917 µs]
                        change: [−4.8868% −4.7097% −4.5320%] (p = 0.07 > 0.05)
                        No change in performance detected.

send+recv 32 KiB        time:   [27.269 µs 27.313 µs 27.324 µs]
                        change: [−8.6706% −6.9166% −5.0977%] (p = 0.05 > 0.05)
                        No change in performance detected.

send+recv 64 KiB        time:   [44.328 µs 44.481 µs 45.093 µs]
                        change: [−6.2649% −4.1514% −1.9787%] (p = 0.08 > 0.05)
                        No change in performance detected.

send+recv 128 KiB       time:   [64.124 µs 66.113 µs 66.610 µs]
                        change: [−11.585% −8.2742% −4.8436%] (p = 0.07 > 0.05)
                        No change in performance detected.

send+recv 256 KiB       time:   [105.51 µs 106.04 µs 108.17 µs]
                        change: [−13.633% −11.488% −9.2904%] (p = 0.07 > 0.05)
                        No change in performance detected.

send+recv 512 KiB       time:   [197.02 µs 198.10 µs 202.42 µs]
                        change: [−18.388% −17.046% −15.697%] (p = 0.07 > 0.05)
                        No change in performance detected.

send+recv 1 MiB         time:   [444.86 µs 449.15 µs 466.29 µs]
                        change: [−8.9748% −5.6687% −2.2826%] (p = 0.10 > 0.05)
                        No change in performance detected.

send+recv 2 MiB         time:   [995.05 µs 1.0024 ms 1.0317 ms]
                        change: [−1.6461% +0.5963% +2.8582%] (p = 0.80 > 0.05)
                        No change in performance detected.

send+recv 4 MiB         time:   [1.7360 ms 1.7431 ms 1.7715 ms]
                        change: [−36.987% −36.283% −35.577%] (p = 0.07 > 0.05)
                        No change in performance detected.

send+recv 8 MiB         time:   [5.5704 ms 5.6127 ms 5.7818 ms]
                        change: [−35.835% −34.044% −32.222%] (p = 0.07 > 0.05)
                        No change in performance detected.

send+recv 16 MiB        time:   [13.193 ms 13.256 ms 13.509 ms]
                        change: [−58.905% −57.314% −55.638%] (p = 0.07 > 0.05)
                        No change in performance detected.

send+recv 32 MiB        time:   [46.068 ms 46.688 ms 46.843 ms]
                        change: [−49.800% −48.207% −46.539%] (p = 0.07 > 0.05)
                        No change in performance detected.

send+recv 64 MiB        time:   [89.254 ms 92.471 ms 93.275 ms]
                        change: [−68.854% −67.041% −65.095%] (p = 0.07 > 0.05)
                        No change in performance detected.

send+recv 128 MiB       time:   [172.57 ms 172.89 ms 174.18 ms]
                        change: [−82.532% −82.294% −82.051%] (p = 0.07 > 0.05)
                        No change in performance detected.

send+recv 256 MiB       time:   [319.75 ms 321.78 ms 322.28 ms]
                        change: [−91.122% −90.788% −90.430%] (p = 0.05 < 0.05)
                        Performance has improved.

send+recv 512 MiB       time:   [621.06 ms 621.64 ms 623.94 ms]
                        change: [−93.734% −93.719% −93.705%] (p = 0.05 < 0.05)
                        Performance has improved.

send+recv 1 GiB         time:   [1.1985 s 1.2037 s 1.2246 s]
                        change: [−96.590% −96.553% −96.516%] (p = 0.00 < 0.05)
                        Performance has improved.
```

You have to love that criterion performance improvement detection logic :sweat_smile: 
> change: [−82.532% −82.294% −82.051%] (p = 0.07 > 0.05)
> No change in performance detected.

XavDesbordes:
Basically, it should dramatically close the performance gap with fastwebsockets, which is good

alexheretic:
critcmp using #497 e2e benches
```
group                master                                 #496
-----                -                                      -
send+recv/1 GiB      29.06     35.2±0.00s    58.1 MB/sec    1.00   1212.3±2.19ms  1689.4 MB/sec
send+recv/128 MiB    5.66  1012.8±65.35ms   252.8 MB/sec    1.00    179.0±0.85ms  1429.9 MB/sec
send+recv/16 MiB     2.39     30.4±0.63ms  1052.6 MB/sec    1.00     12.7±0.32ms     2.5 GB/sec
send+recv/2 MiB      1.01   948.1±23.03µs     4.1 GB/sec    1.00   935.2±21.91µs     4.2 GB/sec
send+recv/256 KiB    1.09    125.7±2.52µs     3.9 GB/sec    1.00    115.2±2.30µs     4.2 GB/sec
send+recv/32 KiB     1.01     29.8±0.61µs     2.1 GB/sec    1.00     29.3±0.50µs     2.1 GB/sec
send+recv/4 KiB      1.05     16.2±0.25µs   483.4 MB/sec    1.00     15.4±0.02µs   506.5 MB/sec
send+recv/512 B      1.02     13.6±0.51µs    72.0 MB/sec    1.00     13.3±0.24µs    73.4 MB/sec
```

XavDesbordes:
That's crazy! well done alex

src/protocol/frame/mod.rs
daniel-abramov: Got it, thanks for explaining!

I thought a bit about it, and correct me if I'm wrong, but this would also mean that when we read from a stream that can return `WouldBlock` (like in the case of `tokio-tungstenite`), we're still going to over-reserve because we'll potentially re-enter the `read_frame()` function multiple times essentially resetting the flag (`WoudBlock` may be returned by `read_in` in real-life but would not not happen in benchmarks due to the use of blocking sockets)?

If my assumption is correct, we could replace `if !reserved_full_msg_len` with something like `if self.in_buffer.capacity() < (len + header.size())` (so we'll eliminate the need for the flag while still ensuring that we don't over-reserve much more than needed).

alexheretic: Yes `WouldBlock` during a large read will indeed currently cause over-reservation. But a capacity check won't help, that's just what reserve already does. If we've written 1MB of a 2MB message then hit WouldBlock the capacity could be 1MB and be enough. We'd want to avoid reserving 2MB in this case.

The code in this PR achieves this for error-less reads. To solve this for `WouldBlock` scenario we should store the flag in `self.header` or just do the reserve call when the header is populated.

alexheretic: https://github.com/snapview/tungstenite-rs/pull/501

PR #515, opened 2025-09-08
Refactor WebSocketContext::read_message_frame

akonradi-signal:
Reduce the amount of nested code by returning earlier on errors and from self-contained logic instead of branching with if/else and match expressions. Use an enum type to consolidate common code and avoid a panic! call.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -673,44 +673,50 @@ impl WebSocketContext {
             }
 
             OpCode::Data(data) => {
+                enum FrameType {
+                    Continue,
+                    Initial(IncompleteMessageType),
+                }
+
                 let fin = frame.header().is_final;
-                match data {
-                    OpData::Continue => {
-                        let mut msg = self
+                let frame_type = match data {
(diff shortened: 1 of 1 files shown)


daniel-abramov:
Superseded by:
- https://github.com/snapview/tungstenite-rs/pull/517
- https://github.com/snapview/tungstenite-rs/pull/518

akonradi-signal:
I'm working on adding support for the ["permessage-deflate" extension](https://datatracker.ietf.org/doc/html/rfc7692) (another take on [Add `permessage-deflate` support, again #426](https://github.com/snapview/tungstenite-rs/pull/426)), which makes this code a bit less compact, and wanted to try to get ahead of that.

Closing the loop: https://github.com/signalapp/tungstenite-rs supports `permessage-deflate`. I incorporated the refactors from #518 as https://github.com/snapview/tungstenite-rs/commit/119f4d7eee0d5c4920a63a0758c973dc7149b34f. I had been worried the would be difficult to adopt but I'm pleasantly surprised with how the overall flow turned out!

PR #520, opened 2025-10-19
refactor(framecodec): reduce zero-fill (memset(0)) to speed up read handling

SongXiaoXi:
## Summary
This PR removes per-read zero-filling in `FrameCodec` by replacing the `resize(.., 0) + truncate(..)` pattern with a direct read into the buffer’s spare capacity, followed by `advance_mut(n)`. This keeps observable behavior the same (success/EOF/error), while eliminating unnecessary `memset(0)` on the hot path.

## Why
On targets where memset is relatively slow (e.g., x86_64-unknown-linux-musl), zero-filling dominates small I/O operations. Avoiding it improves CPU efficiency without changing semantics.

## Benchmark (e2e, x86_64-unknown-linux-musl)
```
send+recv/512 B         time:   [4.1185 µs 4.1723 µs 4.2298 µs]
                        thrpt:  [230.88 MiB/s 234.06 MiB/s 237.12 MiB/s]
                 change:
                        time:   [−8.4285% −6.5048% −4.6792%] (p = 0.00 < 0.05)
                        thrpt:  [+4.9089% +6.9574% +9.2043%]
                        Performance has improved.

```

### DIFF
--- a/src/protocol/frame/mod.rs
+++ b/src/protocol/frame/mod.rs
@@ -17,7 +17,7 @@ use crate::{
     protocol::frame::mask::apply_mask,
     Message,
 };
-use bytes::BytesMut;
+use bytes::{BufMut, BytesMut};
 use log::*;
 use std::io::{self, Cursor, Error as IoError, ErrorKind as IoErrorKind, Read, Write};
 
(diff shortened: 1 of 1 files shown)


daniel-abramov:
Thanks for the PR, but unfortunately, this implementation is unsound as it may lead to undefined behavior, see [`Read::read`](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read) (the last two paragraphs).

Also https://github.com/snapview/tungstenite-rs/pull/471#issuecomment-2554315848 might be useful for context.

Issue #523, opened 2025-11-23
Is there any way at all to send data between recieving a close frame and sending the reply close frame?


Scripter17:
I'm making a thing that takes lines of tasks in Message::Text and returns lines of results in Message::Text.

For performance, I want to send response lines in fewer messages by accumulating them in a buffer and sending that buffer on some interval. This makes my code anywhere from 5 to 7 times faster.

Basically within a [`rocket_ws::WebSocket::channel`](https://docs.rs/rocket_ws/latest/rocket_ws/struct.WebSocket.html#method.channel) (which uses tungstenite internally) I have

1. A `tokio::spawn`'d future that repeatedly waits 1ms, gets all available results, and if there's any to return, returns them all at once.
2. A loop that repeatedly gets the next message, then for each line of text parses the task, does it, and sends it over to the responding thread.

The problem is that the instant I read a `Message::Close` from the stream it is impossible to send any further responses because tungstenite immediately closes the connection with no consideration that I might actually have shit to do before closing the connection.

If I do it without the responding future, sending all the results from each message before getting the next message, it reliably sends all responses at the cost of taking 5 to 7 times as long.

If I were to manually reimplement the relevant parts of rocket_ws on top of tungstenite, is there any way at all for me to get the 5-7x performance boost without having a race condition forced upon me? I don't care how low level I have to go I just want this over with.

Scripter17:
I'd like to apologize for my tone throughout this issue. As much as this personally annoys me and as much as I have tried to show the annoyance in somewhat not rude ways I am definitely coming off a lot angrier than I should to people with much more important concerns than letting me do library/protocol level UB for situations even I admit are questionable and performance even I doubt the existence of.

I maintain that Tungstenite should allow an "I'm not stupid" mode for people who do genuinely need this for compatibility or whatever, but I let myself pretend I was coming off as firm and direct when I was just coming off as an ass, and that's not a good way to request a feature.

If you are willing to accept such a feature, I am willing to have a go at adding fields to WebSocketConfig for disabling auto-pong and auto-confirm-close (with both defaulting to enabled). I probably won't do a great job on my first try and it'll probably take a while to even finish that first try, but it'd much more productive than what I've been doing above.

220nightmore-spec:
[...]

PR #527, opened 2026-01-02
Implement WebSocketConfig::manually_close (#523)

Scripter17:
The docs almost certainly need improvement.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -90,6 +90,9 @@ pub struct WebSocketConfig {
     /// some popular libraries that are sending unmasked frames, ignoring the RFC.
     /// By default this option is set to `false`, i.e. according to RFC 6455.
     pub accept_unmasked_frames: bool,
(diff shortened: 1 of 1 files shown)


Scripter17:
> Just curious about the comment
> 
> ```
>     /// When set to `true`, receiving a close message won't automatically close the connection.
>     /// By default this option is set to `false`.
> ```
> 
> When is that necessary and how is the connection closed instead in that case?

For me specifically, I want to be able to process and respond to messages in parallel and only respond to the close message when I'm done sending results. Currently I am processing each task in a message in parallel, waiting for all the results to be sent, and only then getting the next message. Being able to disable the automatic closing would make my code much simpler and (theoretically) a bit faster.

To close the connection, you just send a close message like any other message. How exactly you store the close message content to send back is up to you.

I'm aware whether or not the connection stays open after the close handshake starts is undefined, but at least WebSocat does the sensible option of waiting for the response.

chtpl:
Ah, OK

Issue #522, opened 2025-11-18
Consider migrating to `rustls-platform-verifier` from `rustls-native-certs`


Enduriel:
[`rustls-platform-verifier`](https://github.com/rustls/rustls-platform-verifier) is recommended over [`rustls-native-certs`](https://github.com/rustls/rustls-native-certs) by the team maintaining `rustls-native-certs`. From the readme:
> Instead of this crate, we suggest using [rustls-platform-verifier](https://github.com/rustls/rustls-platform-verifier), which provides a more robust solution with a simpler API. This crate is still maintained, but mostly for use inside the platform verifier on platforms where no other solution is available. For more context, see [deployment considerations](https://github.com/rustls/rustls-platform-verifier?tab=readme-ov-file#deployment-considerations).

 Migrating would also allow using the native-tls feature on android/ios, which are not supported by `rustls-native-certs`.

Is this something the tungstenite team would be interested in doing/receiving a PR for?

TitanNano:
I just had the issue that `rutls-native-certs` for some reason (I wasn't able to determine the exact cause) is not able to pick up our corporate intranet CA. With [`rustls-platform-verifier`](https://github.com/rustls/rustls-platform-verifier) on the other hand, the certificates are handled correctly. 

It is currently possible to switch to using the `rustls-platform-verifier` by directly calling `client_tls_with_config` but this means you have to effectively duplicate `connect_with_config`.

daniel-abramov:
Interesting. Did you have a chance to debug what causes this behavior? (it's obviously not `tungstenite` related, but it would certainly be interesting to know why the two differ and if there are ways to make it properly work)

(I believe you would be able to reproduce it without tungstenite by using the `rustls-native-certs` and `rustls-platform-verified` directly)

TitanNano:
> Did you have a chance to debug what causes this behavior?

Unfortunately, no. I forgot to mention that this occurred on macOS. 

I spent a day debugging `rutls-native-certs`. It makes a call to an Apple system API to fetch the root certificates, which returns a list. This list is supposed to contain SYSTEM, ADMIN, and USER certificates, but I was unable to locate our intranet root certificate. [Apple API](https://developer.apple.com/documentation/security/sectrustsettingscopycertificates(_:_:))

[...]

PR #528, opened 2026-01-02
Add support for rustls-platform-verifier

FrankenApps:
Adds support for [`rustls-platform-verifier`](https://github.com/rustls/rustls-platform-verifier) as mentioned in #522.

However contrary to the plan outlined in that issue I left the [`rustls-native-certs`](https://github.com/rustls/rustls-native-certs) feature untouched for now.

I also did not implement support for additional root certificates ([`new_with_extra_roots`](https://docs.rs/rustls-platform-verifier/latest/rustls_platform_verifier/struct.Verifier.html#method.new_with_extra_roots)), because it works differently than the current approach based on [`RootCertStore`](https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ native-tls = ["native-tls-crate"]
 native-tls-vendored = ["native-tls", "native-tls-crate/vendored"]
 rustls-tls-native-roots = ["__rustls-tls", "rustls-native-certs"]
 rustls-tls-webpki-roots = ["__rustls-tls", "webpki-roots"]
+rustls-tls-native-platform-verifier = ["__rustls-tls", "rustls-platform-verifier"]
 __rustls-tls = ["rustls", "rustls-pki-types"]
 
 [dependencies]
@@ -60,6 +61,10 @@ version = "0.8.0"
 [dependencies.webpki-roots]
 optional = true
 version = "0.26"
+
+[dependencies.rustls-platform-verifier]
+optional = true
+version = "0.6"
 
 [dev-dependencies]
 criterion = "0.6"
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -70,7 +70,9 @@ mod encryption {
 
     #[cfg(feature = "__rustls-tls")]
     pub mod rustls {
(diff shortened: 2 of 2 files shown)


daniel-abramov (commented):
Thanks for the PR!

> @daniel-abramov Is this something you would consider?

Yes. I think it's the right direction in the long term.

The only concern I have with these particular changes is that cargo features are meant to be additive, and I think it could be seen as an unexpected behavior if one feature "silently overwrites" the other.

> CI failure seems unrelated:

Oh right. I've just updated the `master` to incorporate the fix.

FrankenApps:
> The only concern I have with these particular changes is that cargo features are meant to be additive, and I think it could be seen as an unexpected behavior if one feature "silently overwrites" the other.

I agree that this is somewhat unfortunate, albeit also expected behavior for these `rustls-tls-*` features.
Most other projects do this in a very similar manner as far as I can tell: [`quinn`](https://github.com/quinn-rs/quinn/blob/a0ec97a50bdab237450dce97b69a63419ef0ef25/quinn-proto/Cargo.toml#L31), [`ureq`](https://github.com/algesten/ureq/blob/08785cb67f3f140d112f00b0e1bf66c46d0cd9f3/Cargo.toml#L27), etc.

A notable exception is the [latest version of `reqwest`](https://github.com/seanmonstar/reqwest/blob/fa74a8b835b2f194253cebdab8d049c0e5d23af3/Cargo.toml#L46) which offers a `rustls-no-provider` feature that at least allows for selecting either `aws-lc-rs` or the `ring` crypto provider, but it still does not allow for additively mixing different TLS verifiers as far as I can tell.  
Because of the way these behave I would even go as far and doubt that it is possible to mix those in any meaningful / logical way.

Therefore while I think that a `rustls-no-provider` feature would likely make sense for `tungstenite-rs` too (might start a separate PR if desired), I personally do not really see how to resolve your concern mentioned above.

If you see it as a blocker for merging and nobody else can give me guidance for how to improve the PR in that regard, feel free to close it.

FrankenApps:
@daniel-abramov Rebased on top of master in order to fix the unrelated clippy issue in CI.

elichai:
Any updates on this?

PR #542, opened 2026-04-22
adding VecBinary to protocol::message:Message

whilesoftware:


daniel-abramov:
The PR has been in a draft state for quite some time now, so I'm gonna close it as stalled.

Having vector writes is goood, but I don't think that the suggested changes of adding an additional `Message` type that changes the implementation details of the crate is not the right mental mode long-term.

See [performance section](https://github.com/snapview/tokio-tungstenite#is-it-performant). The right approach would probably be to have a `tungstenite-core` that is a sans-io websocket state machine that does no own I/O at all. And then on top of that, build the sync version of `tungstenite` and `tokio-tungstenite`. This would make a much cleaner split, with many more opportunities to optimize performance beyond vectored writes (e.g., our current layer of buffered reads/writes could be rewritten in a much more elegant way once the I/O and the state machine have a clean split).

PR #544, opened 2026-05-20
chore: improve tungstenite-rs maintenance path

oab24413gmai:
Summary:
- Add targeted unit tests in src/protocol/frame/coding.rs for OpCode/CloseCode round-trip conversions (From<u8>/Into<u8>) covering reserved/invalid byte ranges and the boundary between Control and Data opcodes, plus a test asserting that CloseCode::from(u16) maps the IANA-reserved/unassigned bands to the correct Reserved/Iana/Library/Bad variants. These are small, self-contained protocol-correctness tests against an existing coding module — no behavior changes required.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/src/protocol/frame/coding.rs
+++ b/src/protocol/frame/coding.rs
@@ -261,6 +261,7 @@ impl From<u16> for CloseCode {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use super::CloseCode::*;
 
     #[test]
     fn opcode_from_u8() {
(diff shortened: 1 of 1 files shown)


daniel-abramov (commented):
While the changes are safe, I don't quite understand the underlying reason for adding them. Do you plan to change or extend these parts? Or did you experience a bug that is related to the newly introduced tests?

I'm asking because the tests add many assertions without much increase in abstraction or safety. They mostly mirror a trivial `match`/`From` mapping and restate implementation details. I'd say we either need (a) good comprehensive tests that check useful invariants or (b) drop them to avoid low-signal test maintenance overhead.

At the moment, I'm not sure the added maintenance costs are justified - we already have more comprehensive e2e tests that cover common websocket flows (and, where applicable, transitively use the conversions).

PR #545, opened 2026-05-26
check frame length against max_size before narrowing to usize

metsw24-max:
The payload length is narrowed to usize before the max_size check, so on 32-bit/wasm32 a length above 2^32 wraps past it; compare in u64 first.

### DIFF
--- a/src/protocol/frame/mod.rs
+++ b/src/protocol/frame/mod.rs
@@ -373,7 +373,7 @@ mod tests {
         let mut sock = FrameSocket::new(raw);
         assert!(matches!(
             sock.read(None),
-            Err(Error::Capacity(CapacityError::MessageTooLong { .. }))
+            Err(Error::Capacity(CapacityError::MessageTooLong { size: 5, max_size: usize::MAX }))
         ));
     }
 }
--- a/src/protocol/frame/mod.rs
+++ b/src/protocol/frame/mod.rs
@@ -175,15 +175,16 @@ impl FrameCodec {
                 bytes::Buf::advance(&mut self.in_buffer, advanced as _);
 
                 if let Some((_, len)) = &self.header {
-                    let len = *len as usize;
-
-                    // Enforce frame size limit early
-                    if len > max_size {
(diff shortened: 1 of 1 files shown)


src/protocol/frame/mod.rs
```diff
@@ -362,4 +363,17 @@ mod tests {
...
+    #[test]
+    #[cfg(target_pointer_width = "32")]
+    fn length_above_usize_max_rejected() {
+        // 64-bit payload length 0x1_0000_0005 does not fit a 32-bit usize; it
+        // must be rejected rather than wrapping to 5 when narrowed.
+        let raw = Cursor::new(vec![0x82, 0x7f, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x05]);
+        let mut sock = FrameSocket::new(raw);
+        assert!(matches!(
+            sock.read(None),
+            Err(Error::Capacity(CapacityError::MessageTooLong { .. }))
```
daniel-abramov: Was `{ ... }` intentional? I.e., the length would be truncated (the rest of the unit tests have real values, precisely for this case, to see what's gonna be returned).

daniel-abramov (commented):
In essence, the changes look correct: we used to compare to `u64` [prior to the changes](https://github.com/snapview/tungstenite-rs/commit/6c5aeb2418bc0a08c856be5759eb0c2661c8eba0) done 2 years ago.

Note that `tungstenite` is unlikely to compile as-is on one of the targets mentioned in the comment (wasm32-unknown-unknown) without minor changes in the dependencies configuration.

metsw24-max:
No, `{ .. }` was not intentional — thanks for catching that. Updated to `{ size: 5, max_size: usize::MAX }` (the truncated low 32 bits of `0x1_0000_0005`, and `usize::MAX` from the `None` default).

Noted on `wasm32-unknown-unknown` — understood that the dependency configuration would need separate work to make it compile there; the test gating on `target_pointer_width = "32"` was intended to cover native 32-bit targets for now.

PR #546, opened 2026-05-27
reject non-minimal frame payload length encoding (RFC 6455 §5.2)

metsw24-max:
FrameHeader::parse accepts a length encoded with the 16/64-bit form when a shorter form would fit, but RFC 6455 §5.2 mandates the minimal encoding.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -256,6 +256,9 @@ pub enum ProtocolError {
     /// The payload for the closing frame is invalid.
     #[error("Invalid close sequence")]
     InvalidCloseSequence,
+    /// The payload length was not encoded with the minimal number of bytes (RFC 6455 §5.2).
+    #[error("Frame payload length not minimally encoded")]
+    NonMinimalLengthEncoding,
 }
 
 /// Indicates the specific type/cause of URL error.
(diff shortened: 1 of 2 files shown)


daniel-abramov:
RFC 6455 §5.2 describes the format (and we do format all the frames in accordance with the RFC). It is slightly less direct about receiver behavior, though.

In practice, the Autobahn already verifies that we do format the frame length correctly, so are all compliant implementations. There is no explicit Autobahn [test](https://github.com/crossbario/autobahn-testsuite/issues/108) verifying that the **receiver** must reject such frames, and many popular WebSocket libraries (including `tunsgtenite`) do not reject them (which may be due to historical implementations that did not strictly follow it).

My broader concern is that I do not understand the purpose of this PR.

What was the rationale behind the additional strictness added to the receiving side? Why is it so strict on one part of the RFC section while ignoring the other part of the same section related to the 64-bit MSB rule? Why do unit tests not cover the other edge cases related to this case? Why does the fix reintroduce magic numbers in-place instead of relying on `LengthFormat` that already deals with those specifics?

Following the same concerns expressed [in the other comment](https://github.com/snapview/tungstenite-rs/pull/550#issuecomment-4603124870), I'd close this one.

PR #548, opened 2026-05-31
split Connection response header into tokens before matching Upgrade

metsw24-max:
A conformant server may send the Connection header as a token list such as `keep-alive, Upgrade`, which verify_response rejected because it matched against the whole value. Split on commas and spaces and match a single token, mirroring the server-side check already in create_parts.

### DIFF
--- a/src/handshake/client.rs
+++ b/src/handshake/client.rs
@@ -245,7 +245,7 @@ impl VerifyData {
         if !headers
             .get("Connection")
             .and_then(|h| h.to_str().ok())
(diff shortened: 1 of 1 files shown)


daniel-abramov:
> A conformant server may send the Connection header as a token list such as keep-alive, Upgrade

A conformant **WebSocket server** must send `Connection: Upgrade`.

> Split on commas and spaces and match a single token, mirroring the server-side check already in create_parts.

This asymmetry arises because some clients **may** send a `Connection` containing a list of values. But the server's reply to the handshake is `Connection: websocket`.

PR #549, opened 2026-06-01
match Upgrade header token-wise like the Connection header

metsw24-max:
The Connection header is matched token-wise (`split([' ', ',']).any(...)`) but the adjacent Upgrade check in `create_parts` and `verify_response` compares the whole value against "websocket". Upgrade is a comma-separated token list (RFC 7230 6.7), so a peer sending e.g. `Upgrade: websocket, h2c` is rejected with MissingUpgradeWebSocketHeader. Split into tokens before matching, same as the Connection check right next to it.

### DIFF
--- a/src/handshake/client.rs
+++ b/src/handshake/client.rs
@@ -233,7 +233,7 @@ impl VerifyData {
         if !headers
             .get("Upgrade")
             .and_then(|h| h.to_str().ok())
-            .map(|h| h.eq_ignore_ascii_case("websocket"))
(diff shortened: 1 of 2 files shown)


daniel-abramov:
Syntactically: Yes, the `Upgrade` is a comma-separated list of tokens.

Semantically: `Upgrade` is not supposed to contain other protocols **in the context of the WebSocket handshake**. The WebSocket server must reply with `Upgrade: websocket`.

PR #550, opened 2026-06-02
fail client handshake on unrequested Sec-WebSocket-Extensions

metsw24-max:
verify_response left RFC 6455 4.1 step 5 as a TODO and never inspected the server's Sec-WebSocket-Extensions header. The generated client handshake offers no extensions, so anything the server echoes there was not requested and the connection must fail. Currently the handshake completes anyway and the peers desync: the server thinks e.g. permessage-deflate is active while the client applies none. Reject it in verify_response, next to the existing accept-key check.

### DIFF
--- a/src/handshake/client.rs
+++ b/src/handshake/client.rs
@@ -262,7 +262,14 @@ impl VerifyData {
         // that was not present in the client's handshake (the server has
         // indicated an extension not requested by the client), the client
         // MUST _Fail the WebSocket Connection_. (RFC 6455)
-        // TODO
+        //
+        // No extensions are offered in the generated client handshake, so any
+        // extension echoed back by the server was not requested.
+        if headers.contains_key("Sec-WebSocket-Extensions") {
+            return Err(Error::Protocol(ProtocolError::InvalidHeader(
+                HeaderName::from_static("sec-websocket-extensions").into(),
+            )));
+        }
 
         // 6.  If the response includes a |Sec-WebSocket-Protocol| header field
         // and this header field indicates the use of a subprotocol that was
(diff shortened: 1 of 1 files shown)


daniel-abramov:
Right, that's why we have a TODO in place. What I can't understand is what problem is this PR supposed to fix. I'd totally understand if this change was made in the context of, e.g., adding support for subprotocols or similar (something we don't support atm).

> Currently the handshake completes anyway and the peers desync: the server thinks e.g. permessage-deflate is active while the client applies none.

This isn't entirely correct. The proper server would only reply with permessage-deflate if the client offered it. `tungstenite` clients never offer permessage-deflate as of now. It's only a malfunctioning / non-compliant server that would reply with an extension that has not been negotiated. But even in such a case, the client would reject a frame once it receives it because of the `rsv` set on per-message-deflate frames.

---

There is a general remark that I'd like to make: your account has submitted several PRs in the past few days. Those PRs make an impression of low-effort AI-generated PRs: i.e., PRs for the sake of PRs, that randomly peek at different parts of code without a clear intention or agenda.

We welcome PRs, but each PR takes time to review - this is justifiable if they bring value and/or were submitted in good faith, trying to improve/fix concrete issues. But "AI fuzzing" causes unnecessary churn without bringing real value.

metsw24-max:
Fair enough. You're right that a compliant server won't echo an extension the client never offered, and the rsv check already catches the malfunctioning case before any state desync matters, so this is really just spec-completeness for something that can't bite today. Fine to leave the TODO until extensions/subprotocols are actually on the table.

On the volume, that's fair too. I'll slow down and keep submissions to concrete issues rather than spec nits. Sorry for the churn.

PR #551, opened 2026-06-04
drop credentials when connect follows a cross-origin redirect

metsw24-max:
**Credentials forwarded across origins on redirect**

`connect` follows redirects by replaying the original request on every hop, so an `Authorization` or `Cookie` set on the initial request is sent to whatever host the server puts in `Location`. A malicious or compromised endpoint can redirect to an origin it shouldn't have the credentials for and harvest them, so I drop those headers (and `Proxy-Authorization`) when the redirect target's scheme, host or port differ from the current one. Regression test in tests/redirect_credentials.rs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# UNRELEASED
+
+* Drop sensitive headers (`Authorization`, `Cookie`, `Proxy-Authorization`, ...) when `connect` follows a redirect to a different origin.
+
 # 0.30.0
 
 * Reject non-compliant clients (incorrect `Sec-WebSocket-Key` length/format) on the server side.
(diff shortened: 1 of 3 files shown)


daniel-abramov:
There is a good point that removing sensitive headers during redirects might be a sensible approach and some libraries follow this rule.

I'm not sure how relevant it is for WebSocket. To my knowledge, many libraries don't even allow redirects, and our current handling of them is bare-bones and maybe needs a revisit if we want to keep it long-term. E.g. the changes that you submitted seem like a good direction, but they are also incomplete compared to e.g. handling of the same redirects [inside `reqwest`.](https://docs.rs/reqwest/latest/src/reqwest/redirect.rs.html#239-252)

PR #552, opened 2026-06-05
reject oversized control frames in frame parser

metsw24-max:
**Oversized control frames buffered before rejection**

RFC 6455 §5.5 caps control frame payloads at 125 bytes, but `FrameHeader::parse_internal` only checks the opcode, not the declared length, so a control-frame header announcing a larger payload is parsed as valid. The codec then reserves and reads the whole declared length (up to `max_frame_size`, 16 MiB by default) before `read_message_frame` rejects it with `ControlFrameTooBig`, meaning a 4-byte header can trigger a multi-megabyte buffer reservation for a frame that was never admissible.

The check now lives next to the existing opcode validation in the parser, so an oversized control frame is rejected from its declared length before any payload is reserved. The later check in `read_message_frame` is left in place as a backstop. This should also keep the length rule easier to reason about, since both control-frame constraints that are knowable from the header are now enforced at parse time.

### DIFF
--- a/src/protocol/frame/frame.rs
+++ b/src/protocol/frame/frame.rs
@@ -197,6 +197,14 @@ impl FrameHeader {
             _ => (),
         }
 
+        // All control frames MUST have a payload length of 125 bytes or less
+        // (RFC 6455, section 5.5). Reject an oversized control frame from its
+        // declared length here, before the codec reserves and buffers the
+        // payload (up to `max_frame_size`).
+        if matches!(opcode, OpCode::Control(_)) && length > 125 {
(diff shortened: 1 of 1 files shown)


daniel-abramov:
As you have already noted, the check in question is already part of the logic, it's just performed a bit later when `match`ing the `opcode`. If the concern was really about sparing that allocation upfront for invalid control frames, then why did you move only the payload check, but not the adjacent fragmentation check that is also known at header parse time?

> The later check in read_message_frame is left in place as a backstop.

??

PR #554, opened 2026-06-13
Replace runtime rand dependency with getrandom

sheawinkler:
## Summary

This removes the runtime `rand` dependency from tungstenite and uses `getrandom` directly for the two protocol entropy call sites:

- WebSocket client masking keys (`[u8; 4]`)
- `Sec-WebSocket-Key` generation during client handshakes (`[u8; 16]`)

The generated byte lengths and public API stay unchanged. The existing `rand` dev-dependency remains in place for benchmarks.

I used `getrandom = "0.3"` so the change stays within the repository's Rust 1.71 MSRV while avoiding the heavier runtime `rand` dependency for these byte-fill operations.

## Tradeoffs

The current APIs are infallible, so this keeps that shape and treats system entropy failure as unrecoverable with explicit `expect(...)` messages. Changing these call sites to return `Result` would be a wider public API change and does not seem warranted for this dependency cleanup.

## Verification

- `RUSTUP_HOME=/Volumes/wd_black/rustup-home CARGO_HOME=/Volumes/wd_black/cargo-home rustup run nightly cargo fmt --all --check`
- `CARGO_TARGET_DIR=/Volumes/wd_black/cargo-targets-upstream/tungstenite cargo clippy --all --tests --all-features -- -D warnings`
- `CARGO_TARGET_DIR=/Volumes/wd_black/cargo-targets-upstream/tungstenite cargo test --release`
- `CARGO_TARGET_DIR=/Volumes/wd_black/cargo-targets-upstream/tungstenite RUSTUP_HOME=/Volumes/wd_black/rustup-home CARGO_HOME=/Volumes/wd_black/cargo-home rustup run 1.71.1 cargo check`
- `PATH=/Volumes/wd_black/cargo-home/bin:$PATH CARGO_HOME=/Volumes/wd_black/cargo-home CARGO_TARGET_DIR=/Volumes/wd_black/cargo-targets-upstream/tungstenite cargo hack check --feature-powerset --all-targets`

Note: `cargo fmt` reports existing warnings about unstable rustfmt config options, but exits successfully.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ bytes = "1.9.0"
 http = { version = "1.0", optional = true }
 httparse = { version = "1.3.4", optional = true }
 log = "0.4.8"
-rand = "0.9.0"
+getrandom = "0.3"
 sha1 = { version = "0.10", optional = true }
 thiserror = "2.0.7"
 url = { version = "2.1.0", optional = true }
--- a/src/handshake/client.rs
+++ b/src/handshake/client.rs
@@ -329,8 +329,9 @@ impl<'h, 'b: 'h> FromHttparse<httparse::Response<'h, 'b>> for Response {
 pub fn generate_key() -> String {
     // a base64-encoded (see Section 4 of [RFC4648]) value that,
     // when decoded, is 16 bytes in length (RFC 6455)
-    let r: [u8; 16] = rand::random();
-    data_encoding::BASE64.encode(&r)
+    let mut key = [0u8; 16];
+    getrandom::fill(&mut key).expect("websocket handshake key requires system randomness");
+    data_encoding::BASE64.encode(&key)
 }
 
 #[cfg(test)]
--- a/src/protocol/frame/mask.rs
+++ b/src/protocol/frame/mask.rs
@@ -1,7 +1,9 @@
 /// Generate a random frame mask.
 #[inline]
 pub fn generate_mask() -> [u8; 4] {
-    rand::random()
+    let mut mask = [0u8; 4];
+    getrandom::fill(&mut mask).expect("websocket mask generation requires system randomness");
+    mask
 }
 
 /// Mask/unmask a frame.


sheawinkler:
Please review @daniel-abramov

daniel-abramov:
Would you mind elaborating on the purpose of this change?

sheawinkler:
No problem, @daniel-abramov 

1. this keeps the protocol code closer to the WebSocket spec ([RFC 6455](https://www.rfc-editor.org/rfc/rfc6455))
2. The two sites only need cryptographically safe random bytes from the OS: 
 - 4 bytes for frame masks and 
 - 16 bytes for Sec-WebSocket-Key 
 
So, pulling in full rand is extra dependency surface for no real protocol benefit. 
getrandom is the direct primitive, keeps the graph leaner, and preserves identical wire behavior.
 - protocol behavior unchanged -- wss handshake and frames look the same on network
 - simply swap randomness source from rand to direct OS-backed randomness -- `getrandom` (CSPRNG)

### sources
* [getrandom](https://docs.rs/getrandom/latest/getrandom)
* [NIST Reference](https://csrc.nist.gov/pubs/sp/800/90/a/r1/final)
* [wiki](https://en.wikipedia.org/wiki/Cryptographically_secure_pseudorandom_number_generator)

daniel-abramov:
> this keeps the protocol code closer to the WebSocket spec ([RFC 6455](https://www.rfc-editor.org/rfc/rfc6455))

How?

Overall, I have the following concerns/questions:
1. I don't understand what problem this PR tries to solve. Typically, users encounter an issue that motivates them to adjust something in `tungstenite` to solve it or enhance a specific behavior. What problem were you trying to solve that motivated this particular change?
2. The only argument that I see so far is a "leaner dependency graph" (that I'd rather translate to "somewhat smaller crate size due to smaller dependency"). However, I'm just not sure if it's worth the change, provided that `getrandom` is a much lower-level crate and that the performance of making a syscall on each random generation is likely to be much slower than using the random generator created by `rand` (yes, I'm aware that "much slower" is a relative term here since even having a x4 slower performance with `getrandom`, we're still talking about nanoseconds, meaning that this change is barely noticeable by the end user).

PR #556, opened 2026-06-15
Add WebSocket::into_inner_with_read_buffer to recover buffered bytes

### LINKED ISSUE #555: No way to recover the codec read buffer when taking over the raw stream after the handshake
### Problem

`WebSocket::into_inner()` returns only the underlying stream and **discards the frame codec's internal read buffer**. When a caller completes the WebSocket handshake with tungstenite and then wants to take over the *raw* stream — e.g. to bypass framing and relay raw bytes (a tunnel / passthrough mode) — any bytes already pulled into the codec's read buffer are silently lost.

This bites in two situations:

1. **Coalesced post-handshake bytes.** A push-first server frequently writes its `101 Switching Protocols` response **and the first WebSocket frame(s)** in the same TCP segment. During the handshake tungstenite reads that whole segment, so the frame bytes land in the codec's `in_buffer` (via `WebSocket::from_partially_read` / the handshake machinery). `into_inner()` then drops them on the floor.
2. **Unread frame tail.** If some frames have already been read, an unconsumed tail can remain in the read buffer; `into_inner()` discards that too.

### Why it can't be worked around today

The residual lives in `FrameCodec::in_buffer: BytesMut`. The frame layer already exposes it — `FrameSocket::into_inner() -> (Stream, BytesMut)` — but the `WebSocket` path has no equivalent: the codec sits behind the private `WebSocketContext.frame` field, inside the private `WebSocket.context` field, so there's no public accessor reachable from `WebSocket` or `WebSocketContext`. `into_inner()` / `get_ref()` only expose the underlying socket. (Verified against `0.29` and `master`.)

This is the blocker that prevents downstream wrappers (e.g. `tokio-tungstenite`'s `WebSocketStream::into_inner()`, see snapview/tokio-tungstenite) from offering a lossless raw-stream takeover.

### Proposal

Add a consuming accessor that returns the stream **with** its buffered-but-unconsumed bytes:

```rust
impl<Stream> WebSocket<Stream> {
    pub fn into_inner_with_read_buffer(self) -> (Stream, bytes::Bytes);
}
impl WebSocketContext {
    pub fn into_read_buffer(self) -> bytes::Bytes;
}
```

`Bytes` is already re-exported by the crate, and `BytesMut::freeze()` is zero-copy. A caller relaying raw bytes writes the returned buffer to the peer **before** copying further from the stream.

PR: I've opened a small implementation with tests + CHANGELOG.

jeremyjpj0916:
Fixes #555.

### Problem

`WebSocket::into_inner()` returns only the underlying stream and discards the frame codec's internal read buffer. When a caller takes over the raw stream after the handshake (e.g. to relay raw bytes / drop to a plain byte copy), bytes the peer coalesced into the same read as the handshake response — or an unread frame tail — are silently lost. The codec's `in_buffer` is unreachable: it sits behind the private `WebSocketContext.frame` field inside the private `WebSocket.context`, and only `FrameSocket` (not the `WebSocket` path) exposes `into_inner() -> (Stream, BytesMut)`.

### Change

Add a consuming accessor that returns the stream together with its not-yet-consumed read buffer:

- `WebSocket::into_inner_with_read_buffer(self) -> (Stream, Bytes)`
- `WebSocketContext::into_read_buffer(self) -> Bytes` (counterpart for callers driving `WebSocketContext` over their own stream)
- `FrameCodec::into_read_buffer(self) -> BytesMut` (internal, `pub(super)`)

`Bytes` is already re-exported by the crate (`pub use bytes::Bytes`, used by `Message`), and `BytesMut::freeze()` is zero-copy, so this adds no new public dependency and no allocation. `into_inner()` is unchanged.

### Tests

- `into_inner_with_read_buffer_recovers_coalesced_bytes` — frame bytes seeded via `from_partially_read` (modelling a coalesced post-handshake segment) are returned, not dropped.
- `into_inner_with_read_buffer_returns_unread_tail` — after reading the first of two buffered frames, the second remains and is recovered.

Full lib test suite passes (50/50); `cargo fmt` and `cargo clippy --lib` are clean. CHANGELOG updated under UNRELEASED.

### Motivation

This is the upstream blocker for a lossless raw-stream takeover after a WebSocket handshake — e.g. a proxy/gateway tunnel mode that completes the handshake with tungstenite/tokio-tungstenite then bypasses framing to relay raw bytes. A companion PR on snapview/tokio-tungstenite surfaces this on `WebSocketStream`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # UNRELEASED
 
 * Reject non-compliant clients (incorrect `Sec-WebSocket-Key` length/format) on the server side.
+* Add `WebSocket::into_inner_with_read_buffer()` (and `WebSocketContext::into_read_buffer()`) so that a caller taking over the raw stream after the handshake can recover bytes the peer coalesced into the same read as the handshake response, which `into_inner()` discards.
 
 # 0.29.0
 
--- a/src/protocol/frame/mod.rs
+++ b/src/protocol/frame/mod.rs
@@ -148,7 +148,19 @@ impl FrameCodec {
 
     /// Consume the codec, returning the bytes that were read from the stream
     /// into the read buffer but not yet consumed as a frame.
+    ///
+    /// Lossless only at a frame boundary. If a partial frame is in flight (a
+    /// header has been parsed but its payload has not fully arrived, e.g. after
+    /// a mid-frame `WouldBlock`), the header bytes have already been consumed
+    /// out of the read buffer and are not recoverable here — the returned
+    /// buffer then holds only the partial payload. Debug builds assert against
+    /// this misuse.
     pub(super) fn into_read_buffer(self) -> BytesMut {
+        debug_assert!(
+            self.header.is_none(),
+            "into_read_buffer called with a partial frame in flight; the parsed \
+             header bytes have already been consumed and cannot be recovered",
+        );
         self.in_buffer
     }
 
--- a/src/protocol/frame/mod.rs
+++ b/src/protocol/frame/mod.rs
@@ -146,6 +146,12 @@ impl FrameCodec {
         }
     }
 
+    /// Consume the codec, returning the bytes that were read from the stream
+    /// into the read buffer but not yet consumed as a frame.
+    pub(super) fn into_read_buffer(self) -> BytesMut {
+        self.in_buffer
+    }
+
     /// Sets a maximum size for the out buffer.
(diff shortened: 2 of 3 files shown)


jeremyjpj0916:
Companion PR that surfaces this on `tokio-tungstenite`'s `WebSocketStream`: snapview/tokio-tungstenite#380 (draft, depends on this landing + a release). Tracking issue: snapview/tokio-tungstenite#379.

PR #559, opened 2026-07-13
Don't panic on interrupted native-tls TLS handshake (#450)

### LINKED ISSUE #450: Issue with native-tls handshake
Hello,

I got an issue using tungstenite with native-tls feature. Sometimes when my project try to connect to a websocket with tungteniste using native-tls, the connection with client_tls panic with this error : "Bug: TLS handshake not blocked".
It's annoying because I don't know how to catch that kind of panic in my connection retry mechanism.
So my question is why does this code needs to panic when TlsHandshakeError::WouldBlock 
It would be much pleasant if the code will return an error instead of panic. So i can retry to connect with client_tls without getting my project exit with panic.

```
match mode {
                Mode::Plain => Ok(MaybeTlsStream::Plain(socket)),
                Mode::Tls => {
                    let try_connector = tls_connector.map_or_else(TlsConnector::new, Ok);
                    let connector = try_connector.map_err(TlsError::Native)?;
                    let connected = connector.connect(domain, socket);
                    match connected {
                        Err(e) => match e {
                            TlsHandshakeError::Failure(f) => Err(Error::Tls(f.into())),
                            TlsHandshakeError::WouldBlock(_) => {
                                panic!("Bug: TLS handshake not blocked")
                            }
                        },
                        Ok(s) => Ok(MaybeTlsStream::NativeTls(s)),
                    }
                }
            }
```

https://github.com/snapview/tungstenite-rs/blob/master/src/tls.rs

Leonard013:
Fixes #450

## Problem

With the `native-tls` backend, establishing a `wss://` connection over a
non-blocking transport can crash the whole process. When the underlying
`Read + Write` stream returns `WouldBlock` (or a dropped connection) partway
through the TLS handshake, tungstenite hits:

```
thread '...' panicked at src/tls.rs:61:
Bug: TLS handshake not blocked
```

This makes it impossible to drive the client handshake on a non-blocking
socket and retry — the exact use case `HandshakeError::Interrupted` exists for.
The `rustls` backend does **not** have this problem, so the two backends behave
asymmetrically for the same input.

## Root cause

`src/tls.rs`, `encryption::native_tls::wrap_stream` (line 60-62 on `master`):

```rust
TlsHandshakeError::WouldBlock(_) => {
    panic!("Bug: TLS handshake not blocked")
}
```

`native_tls::TlsConnector::connect` drives the TLS handshake *eagerly*, so on a
non-blocking transport it legitimately returns
`native_tls::HandshakeError::WouldBlock(MidHandshakeTlsStream)`. The code
assumed this could never happen and panicked instead of handling it.

The `rustls` backend avoids the whole situation because it never runs the
handshake inside `wrap_stream`: it wraps the socket in a `rustls::StreamOwned`
and lets the TLS handshake complete *lazily* on the first read/write. The
resulting `WouldBlock` then flows through the normal handshake machine and
becomes `HandshakeError::Interrupted`.

## Fix

Mirror the rustls semantics for native-tls, as suggested by @daniel-abramov in
the issue ("wrap the type that native-tls contains so that handling of the TLS
handshake with native-tls remains identical to the end user").

A small wrapper type `stream::NativeTlsStream<S>` now holds either the completed
`native_tls::TlsStream<S>` or the interrupted `MidHandshakeTlsStream<S>`, and
resumes the handshake transparently on the next `read`/`write`:

- `wrap_stream` stores the mid-handshake stream instead of panicking.
- On the next I/O attempt the wrapper calls `MidHandshakeTlsStream::handshake()`.
  If it would block again, it keeps the mid-handshake state and returns
  `io::ErrorKind::WouldBlock`; once complete it transitions to the ready state
  and delegates I/O to the `TlsStream`.

Because the first thing the WebSocket client handshake does is *write* the
upgrade request, that `WouldBlock` is turned by the handshake machine into
`RoundResult::WouldBlock` → `HandshakeError::Interrupted(MidHandshake { .. })` —
**identical to the rustls backend**. The caller can then call `.handshake()`
again to retry, and the resume logic continues the TLS handshake from where it
left off.

Blocking callers are unaffected: `connect` completes eagerly, the wrapper starts
in the `Ready` state, and every I/O call delegates straight to the inner
`TlsStream` (one extra enum branch, no behavioural change).

Files changed:
- `src/tls.rs` — convert the panic into the interrupted/ready paths.
- `src/stream.rs` — add `NativeTlsStream` wrapper (Read/Write/Debug/NoDelay,
  plus `get_ref`/`get_mut`/`into_inner` accessors) and change the
  `MaybeTlsStream::NativeTls` payload to it.
- `tests/native_tls_wouldblock.rs` — regression tests.

## Test

`tests/native_tls_wouldblock.rs` feeds `client_tls` a mock stream whose `read`
always returns `WouldBlock` (writes are accepted), reproducing the reporter's
scenario deterministically. One test asserts the result is
`HandshakeError::Interrupted`; a second asserts `NativeTlsStream::get_ref()`
returns `None` while the handshake is still pending.

**Before the fix** (only `src/` reverted to base, test kept):

```
thread 'native_tls_handshake_wouldblock_is_interrupted_not_panic' panicked at src/tls.rs:61:33:
Bug: TLS handshake not blocked
test native_tls_handshake_wouldblock_is_interrupted_not_panic ... FAILED
test result: FAILED. 0 passed; 1 failed; ...
```

**After the fix:**

```
test native_tls_handshake_wouldblock_is_interrupted_not_panic ... ok
test result: ok. 1 passed; 0 failed; ...
```

## Verification

All commands run on macOS (native-tls = Secure Transport); toolchain 1.96.0,
nightly rustfmt as used by CI.

```
# Full native-tls suite (48 unit tests + integration incl. the regression) — PASS
cargo test --features native-tls
    test result: ok. 48 passed; 0 failed  (lib)
    test native_tls_handshake_wouldblock_is_interrupted_not_panic ... ok

# Default features (no TLS) — PASS (also confirms the fix compiles without native-tls)
cargo test

# Formatting — clean
cargo +nightly fmt --all --check

# Lints, all features (native-tls + rustls together), warnings-as-errors — clean
cargo clippy --all --tests --all-features -- -D warnings

# Feature-gating sanity — both succeed
cargo check --no-default-features
cargo check --features rustls-tls-webpki-roots
```

## Compatibility

This is a **semver-breaking** change to one public type: the payload of
`MaybeTlsStream::NativeTls` changes from `native_tls::TlsStream<S>` to
`stream::NativeTlsStream<S>`. Code that pattern-matches that variant and uses
the inner value as a `native_tls::TlsStream` will need to adapt.
`NativeTlsStream<S>` implements `Read`, `Write`, `Debug` and `NoDelay`, so most
usage through `MaybeTlsStream` is unaffected. To preserve access to the
underlying `native_tls::TlsStream` (e.g. `peer_certificate()`), it also exposes
`get_ref()`, `get_mut()` and `into_inner()`, each returning `Some(..)` once the
handshake has completed and `None` while it is still pending. This matches the
"wrap the type that native-tls contains" direction from the issue. No
`CHANGELOG.md` entry was added since there is no "Unreleased" section yet — glad
to add one under whatever next version you'd like.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -14,7 +14,7 @@ use std::{
 use std::net::TcpStream;
 
 #[cfg(feature = "native-tls")]
-use native_tls_crate::TlsStream;
+use native_tls_crate::{HandshakeError as NativeHandshakeError, MidHandshakeTlsStream, TlsStream};
 #[cfg(feature = "__rustls-tls")]
 use rustls::StreamOwned;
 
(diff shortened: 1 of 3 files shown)


daniel-abramov:
This is way too over-engineered for a rare edge case fix.

Note that #450 is generally rare in any real application because it can only happen when `tungstenite` is used **directly** with non-blocking sockets and `native-tls` is enabled (i.e., it does not affect async users who use `tokio-tungstenite` or those who build on top of `tungstenite` as a WebSocket backend for their own libraries or frameworks).

Issue #557, opened 2026-06-29
Handshake can panic when a custom stream reports zero-byte writes


xc01:
Hi, thanks for maintaining tungstenite.

I found a small robustness issue in the client handshake write path. `tungstenite` is described as a lightweight stream-based WebSocket implementation and a barebone building block for custom networking integrations, and the public `client` API accepts any stream implementing `Read + Write`.

If a custom stream reports `Ok(0)` from `write()` during the handshake write phase, the handshake currently reaches an internal assertion instead of returning a normal transport/handshake error.

Relevant code:

```rust
// src/handshake/machine.rs
HandshakeState::Writing(mut buf) => {
    assert!(buf.has_remaining());
    if let Some(size) = self.stream.write(Buf::chunk(&buf)).no_block()? {
        assert!(size > 0);
        buf.advance(size);
        ...
    }
}
```

Minimal reproducer:

```rust
use std::io::{self, Read, Write};

use tungstenite::client::client;

struct ZeroWriteStream;

impl Read for ZeroWriteStream {
    fn read(&mut self, _buf: &mut [u8]) -> io::Result<usize> {
        Ok(0)
    }
}

impl Write for ZeroWriteStream {
    fn write(&mut self, _buf: &[u8]) -> io::Result<usize> {
        Ok(0)
    }

    fn flush(&mut self) -> io::Result<()> {
        Ok(())
    }
}

#[test]
#[should_panic]
fn client_handshake_panics_on_zero_progress_write() {
    let _ = client("ws://example.com/socket", ZeroWriteStream);
}
```

I understand that returning `Ok(0)` for a non-empty write is unusual for a useful stream. But since tungstenite is often used as a stream-based integration layer, it may be better to treat this as a `WriteZero`/I/O failure or handshake error instead of panicking.

Expected behavior:

The handshake returns an error.

Actual behavior:

The handshake panics at `assert!(size > 0)`.

<details>
<summary>Additional lower-impact panic reachable through <code>IntoClientRequest</code></summary>

I also found another panic path through the public `IntoClientRequest` implementation for `httparse::Request`. I think the impact is probably low because it depends on manually constructing an inconsistent `httparse::Request` rather than using the normal parser-complete path. Still, I am including it here because it is public API reachable and does trigger a panic.

Relevant code:

```rust
// src/client.rs
impl IntoClientRequest for httparse::Request<'_, '_> {
    fn into_client_request(self) -> Result<Request> {
[...]

PR #560, opened 2026-07-25
fix the issue #557.

VinayNishad97:
i returned a std io error if the stream writes 0 bytes . therefor it will not reach the assertion .


youdie006:
Thanks for tackling #557. Two small suggestions that might help this land:

1. A regression test would guard this against future refactors. A minimal deterministic one (no threads/network) is enough: a `ZeroWriteStream` whose `write` always returns `Ok(0)`, then assert that `client("ws://example.com/socket", stream)` returns `Err` instead of panicking.

2. With the early `if size == 0 { return Err(...) }` in place, the following `assert!(size > 0)` is now unreachable and can be dropped.

Happy to share a ready-to-use test if that would be helpful.

